PR #1, opened 2026-08-14
fix(adapters): sanitize connector tool names for OpenAI Codex

victormorenos:
## Problem

Connector-discovered tools can contain characters such as `.`, `/`, whitespace, accents, or Unicode. Pi forwards these names as function tool names, but OpenAI Codex requires names matching `^[a-zA-Z0-9_-]+$` and no longer than 64 characters.

This can fail with an error such as:

```text
[StringParam] [tools[7].name] [invalid_string]
Invalid 'tools[7].name': string does not match pattern.
```

## Fix

- Sanitize only the model-facing tool name at the Pi boundary.
- Preserve the original connector name for execution and dispatch.
- Keep already-valid names unchanged.
- Ensure normalized names are unique and within the provider limit.
- Apply the same boundary to nested subagents.

## Tests

- `pnpm verify:fast`
- `pnpm check`
- `pnpm lint`
- `git diff --check`

The tests cover punctuation, whitespace, Unicode, long names, deterministic collision handling, and protection against shadowing valid tool names.

### DIFF
--- a/packages/adapters/src/pi-runtime.test.ts
+++ b/packages/adapters/src/pi-runtime.test.ts
@@ -1,5 +1,10 @@
+import type { ConnectorTool } from "@rakazo/adapter-kit";
 import { describe, expect, it } from "vitest";
-import { PiAgentRuntime } from "./pi-runtime.js";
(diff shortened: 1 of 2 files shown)


elie222:
Thanks first PR merged!!

PR #2, opened 2026-08-14
fix(adapters): persist OAuth credentials in Pi runtime

victormorenos:
## Problem

Rakazo persists subscription OAuth credentials in its encrypted application store, but the Pi runtime created `builtinModels()` with Pi's default in-memory credential store. After a successful ChatGPT/Codex login, the runtime therefore reported:

```text
Provider is not configured: openai-codex
```

The same boundary affects other Pi OAuth providers such as GitHub Copilot and xAI.

## Fix

- Carry the authorized OAuth credential only in-process through the adapter request.
- Bridge it into Pi with a request-scoped `CredentialStore`.
- Keep refresh publication on the executor's encrypted persistence path.
- Serialize credential refresh/persistence per stored credential.
- Reuse the request-specific model collection for nested agents.
- Preserve the existing API-key path and provider catalog behavior.

No OAuth token or secret is added to an RPC contract, log, or repository file.

## Scope

This PR is limited to `adapter-kit` and the Pi adapter/executor. It does not include Rakazo product UI, model settings, reasoning effort, deployment configuration, or database schema changes.

## Validation

- `pnpm verify:fast` — 149 passed, 18 skipped
- `pnpm check` — 19 successful tasks
- `pnpm lint` — no errors (Biome reports only the existing configuration deprecation notice)
- Added regression tests proving that the persisted OAuth credential configures `openai-codex` and that refresh publication is serialized without exposing credential values in metadata.


elie222:
Merged `main` into this branch and resolved the conflicts.

The jobs rewrite (#5) and sandbox work (#8) had moved `executor.ts` / `pi-runtime.ts` far enough that the original patch no longer applied. The OAuth bug itself was **not** already on main: Pi still used a process-wide `builtinModels()` store, so ChatGPT/Codex (and other OAuth-only providers) could report `Provider is not configured` after a successful login.

What landed here:
- Request-scoped Pi credential store for the current OAuth credential
- Mid-run refresh published back to the encrypted secret store
- Nested agents reuse that same Models collection
- API-key runs stay on the catalog path (they already worked)
- `getApiKey` is omitted on OAuth runs so dual-auth providers (xAI) actually use the store

`pnpm check` and `pnpm verify:fast` passed locally after the merge.

PR #12, opened 2026-08-16
Fix Docker Compose quick start: supervisor workspace build, api EACCES on prisma generate, web missing env

### LINKED ISSUE #11: docker compose up --build fails: supervisor image cannot resolve @rakazo/core@workspace:* (build context excludes the workspace)
## Summary

A fresh clone + the documented Docker Compose quick start fails at the `supervisor` image build. The supervisor's `package.json` declares workspace dependencies, but its Dockerfile builds from an isolated context (`infra/sandboxes/supervisor`) that contains none of the workspace, so pnpm cannot link them.

## Reproduction

```bash
git clone https://github.com/elie222/rakazo.git
cd rakazo
cp .env.example .env   # set real secrets
docker compose -f infra/compose/docker-compose.yml up --build -d
```

## Error

```
target supervisor: failed to solve: process "/bin/sh -c corepack enable && pnpm add tsx hono @hono/node-server dockerode zod pino" did not complete successfully: exit code: 1

[ERR_PNPM_WORKSPACE_PKG_NOT_FOUND] In : "@rakazo/core@workspace:*" is in the dependencies
but no package named "@rakazo/core" is present in the workspace
Packages found in the workspace: @rakazo/sandbox-supervisor
```

## Root cause

- `infra/compose/docker-compose.yml` builds the supervisor with `context: ../sandboxes/supervisor`.
- `infra/sandboxes/supervisor/Dockerfile` copies only that package's `package.json` and runs `pnpm add ...`.
- `pnpm add` also resolves existing dependencies, which include `"@rakazo/core": "workspace:*"` — and `src/index.ts` genuinely imports `resolveSupervisorToken` from it, so the dependency can't simply be dropped.
- `@rakazo/core` in turn depends on `@rakazo/adapter-kit` -> `@rakazo/contracts` (both `workspace:*`), so the image needs the workspace chain.

## Fix that works for me

Build from the repo root with the workspace chain included.

`infra/sandboxes/supervisor/Dockerfile`:

```dockerfile
FROM node:22-bookworm-slim
WORKDIR /app
COPY pnpm-workspace.yaml pnpm-lock.yaml package.json ./
COPY packages/contracts packages/contracts
COPY packages/adapter-kit packages/adapter-kit
COPY packages/core packages/core
COPY infra/sandboxes/supervisor infra/sandboxes/supervisor
RUN corepack enable && pnpm install --filter "@rakazo/sandbox-supervisor..." && pnpm --filter "@rakazo/sandbox-supervisor" add tsx
WORKDIR /app/infra/sandboxes/supervisor
CMD ["pnpm", "exec", "tsx", "src/index.ts"]
```

`infra/compose/docker-compose.yml`:

```yaml
  supervisor:
    build:
      context: ../..
      dockerfile: infra/sandboxes/supervisor/Dockerfile
```

Notes: the root `.dockerignore` already keeps the context lean (`node_modules`, `data`, `.env`, etc. are excluded). Adding `tsx` via a separate `pnpm add` mirrors the old Dockerfile's behavior since it isn't in the supervisor's declared deps — adding it to `devDependencies` upstream would be cleaner. Happy to turn this into a PR.

**Environment:** Debian 13, Docker Engine (compose v2), fresh clone of `main`, 2026-08-16.

Belkouche:
Fresh clones cannot start via the documented `docker compose -f infra/compose/docker-compose.yml up --build` path. This PR fixes the three blockers hit in sequence (full analysis in #11):

1. **supervisor image cannot build** — its `package.json` depends on `@rakazo/core@workspace:*` (genuinely imported by `src/index.ts` for `resolveSupervisorToken`), but the build context was the package directory alone, so pnpm fails with `ERR_PNPM_WORKSPACE_PKG_NOT_FOUND`. The Dockerfile now builds from the repo root and copies the workspace chain (`contracts` → `adapter-kit` → `core` → supervisor), installing with `pnpm install --filter "@rakazo/sandbox-supervisor..."`. The root `.dockerignore` keeps the context lean. (`tsx` is still added explicitly since it is not in the supervisor's declared deps — moving it to `devDependencies` would be a nice follow-up.)

2. **api container exits with EACCES** — the image runs as `USER node` but the compose command re-ran `pnpm --filter @rakazo/db generate` at startup, which tries to rewrite root-owned files under `packages/db/src/generated`. The Prisma client is already generated during the image build, so the runtime regeneration is dropped; `prisma migrate deploy` stays.

3. **web container exits refusing to start** — it depended on `${BETTER_AUTH_SECRET}` interpolation, but Compose v2 reads `.env` from the compose file's directory (`infra/compose/`), not the repo root, so the value was empty and `vite.config.ts` correctly aborted. `web` now uses the same `env_file: ../../.env` as api/worker (which also provides `ENCRYPTION_KEY` and `RAKAZO_HOST` for `preview.allowedHosts`).

Verified end-to-end on Debian 13 / Docker Engine + compose v2: all five services run, `/health` returns `{"ok":true,"runtime":"pi","sandbox":"docker","jobs":"graphile","realtime":"postgres"}`, and the web app serves and signs up correctly from LAN.

Fixes #11


Belkouche:
Pushed a second commit covering a fourth fresh-install blocker found after the first three: Docker creates the `./data` bind-mount directory as **root** on first `compose up`, while api/worker run as `USER node` — so every run fails in ~40 ms writing run state, and the UI sits on "working…" forever. The image actually prepares `/data` with node ownership, but the bind mount shadows it.

Fix: a one-shot `data-init` service (busybox `chown`, using the same `service_completed_successfully` pattern as the existing `computer` service) that api/worker now depend on.

Also included: the bare `catch {}` in `continueRun` rewrote the real exception to `Run setup failed; retrying` across 25 silent retries, which made this genuinely painful to diagnose — it now `console.error`s the underlying error first (matching the package's existing style in `job-reconciler.ts` / `wakeup.ts`).

Verified on the same Debian 13 deployment: runs complete, per-bot computer containers spawn, bot replies stream.

elie222:
Thanks!

PR #62, opened 2026-08-17
Let routines be edited and deleted without deleting the whole bot

### LINKED ISSUE #60: Routines can only be deleted by deleting the whole bot; editing one creates a duplicate instead of updating
The routines API already supports `update` and `remove` (`apps/api/src/router.ts`, `routines.update` / `routines.remove`) — both are fully implemented and functional server-side. But the web app's routine panel (`apps/web/src/pages/Shell.tsx`) never calls either one.

**Repro:**
1. Create a bot with a routine.
2. Open that routine from the list and click Save (even with no changes).
3. A second, duplicate routine appears — Save unconditionally calls `routines.create`, regardless of whether you opened an existing routine or started a new one.
4. There is no delete action anywhere in the UI. The only way to remove a routine is deleting the entire bot, which also destroys its thread, computer, and memory.

**Suggested fix:** track which routine is currently being edited; route Save to `routines.update` when editing an existing one, `routines.create` only for genuinely new routines; add a delete action that calls `routines.remove`.

Happy to open a PR — I have a working fix.

taur-us:
Fixes #60

The routines API already supports `update` and `remove` (`apps/api/src/router.ts`) — both fully implemented and functional server-side. The web routine panel (`apps/web/src/pages/Shell.tsx`) never called either one: Save always called `create`, so opening an existing routine and hitting Save silently created a duplicate rather than updating it, and there was no delete action anywhere in the UI. The only way to remove a routine was deleting the whole bot.

## Changes

- Track which routine is currently being edited (`editingRoutineId`), so Save routes to `routines.update` when editing an existing routine and `routines.create` only for genuinely new ones.
- Add a delete action for routines, using the same confirm-dialog pattern as `DeleteBotDialog` (a lightweight `DeleteRoutineDialog` — no memory-handling choice needed, since routines don't have sub-resources like bots do).

## Testing

- `apps/web` typechecks clean.
- Manually verified in the browser: editing an existing routine and saving no longer creates a duplicate; deleting a routine removes it from the list without touching the bot.


taur-us:
Fixed the active-bot staleness Greptile flagged — pushed as a follow-up commit. The routine panel now resets `editingRoutineId` (and closes if open) whenever the active bot changes, so Save can no longer target a different bot's routine after switching in the sidebar.

elie222:
Thanks!

PR #80, opened 2026-08-18
Add Box sandbox provider

elie222:
## Summary

- add an official ASCII Box sandbox adapter with lifecycle, shell, files, workspace checkpoint/restore, browser streaming, screenshots, and computer actions
- add Box configuration, factory/contracts, canary and Playwright wiring, docs, and deterministic emulator/conformance/fault tests
- enforce Box's single-desktop limitation while keeping shell/files usable in parallel

## Live validation

- real Box canary passed
- targeted Rakazo UI journey passed against a live Box (onboarding, protected input, noVNC takeover, release/resume, routine/plugins/export)
- independent real computer-use pointer click verified against a live Box
- cleaned up all live boxes; no API key is included in the commit

## Verification

- `pnpm lint`
- `pnpm check`
- `pnpm test` (415 passed, 44 skipped)
- `git diff --check`

The Box trial account hit its daily request limit during an extra post-validation rerun; that retry created no box. Pre-existing local Currents/desktop/docs changes remain unstaged and are not part of this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,9 @@ SANDBOX_PROVIDER=docker
 DAYTONA_API_KEY=
 DAYTONA_API_URL=
 DAYTONA_TARGET=
+BOX_API_KEY=
+# Optional Box API base URL; defaults to https://ascii.dev/api/box/v1.
+BOX_API_URL=
 AGENT_RUNTIME=pi
 WAKEUP_DRIVER=graphile
 # Pause or stop computers after this many idle ms. Minimum 30000.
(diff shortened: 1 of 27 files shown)


elie222:
Greptile's noVNC hardening note was reviewed. Box's documented desktop endpoint returns a secret-bearing URL when `publicAccess: false`, but the API/SDK exposes no separate server-bound read-only/control stream or control token; `view_only` is the noVNC client-side mode flag. Rakazo keeps the URL private and gates all server-side `sendInput`/`act` calls through its existing computer-control lease, while the adapter explicitly reports Box's single-desktop limitation. There is no supported Box-side primitive to make this mode immutable, so changing the URL behavior would be misleading.

PR #92, opened 2026-08-19
feat(www): put landing demo bot list behind a hamburger on mobile

elie222:
At phone widths the landing demo stacked the whole bot list above the thread, so the chat was pushed off screen. It now mirrors the mobile app: one screen at a time, with the bot list in a slide-in drawer opened from a hamburger in the thread topbar (closes on select, scrim tap, or Escape) and the computer/settings panel as a full-frame overlay that starts closed.

The `≤860px` layout is now a single column with the sidebar and panel positioned over it; a `matchMedia` listener drives the compact behaviour so hydration stays consistent, and the drawer is `visibility: hidden` when closed to keep it out of the focus order. Desktop layout is unchanged.

Tested with `pnpm --filter @rakazo/www check` and `build`, plus Playwright screenshots at 390px, 820px and 1440px covering drawer open/close, bot select, the computer panel overlay, and resizing between mobile and desktop.

### DIFF
--- a/apps/www/src/components/ProductDemo.tsx
+++ b/apps/www/src/components/ProductDemo.tsx
@@ -342,6 +342,8 @@ export function ProductDemo() {
   const [routineDraft, setRoutineDraft] = useState<RoutineDraft | null>(null);
   const scrollRef = useRef<HTMLDivElement | null>(null);
   const timersRef = useRef<number[]>([]);
+  const menuButtonRef = useRef<HTMLButtonElement | null>(null);
+  const widePanelRef = useRef(true);
   const booting = bootPct > 0;
 
   const active = bots.find((bot) => bot.id === activeId) ?? bots[0];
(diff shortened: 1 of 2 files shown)


elie222:
Both Greptile findings are fixed in ae5504b:

- **P1 desktop panel state** — the compact effect now stores the panel state from the render where the breakpoint flipped and restores it when the viewport grows back, so a round trip no longer strands the desktop frame collapsed (and a panel the user closed at wide widths stays closed).
- **P2 drawer focus** — Escape, the close button, the scrim and bot selection all route through `closeMenu()`, which returns focus to the hamburger; it no-ops when the drawer is already closed so desktop bot clicks keep focus on the row.

Verified in a browser: panel present after desktop → mobile → desktop, still absent after a manual close round trip, and `document.activeElement` is the hamburger after each of the four close paths.

PR #98, opened 2026-08-20
Add Ollama and local MLX server as OpenAI-compatible model providers

taur-us:
## Summary

Adds Ollama and a self-hosted MLX server (e.g. `mlx-openai-server`) as keyless, OpenAI-compatible model providers on top of Pi's built-in catalog, so bots can run entirely on-device with no API key and no network egress. The rest of the commits are fixes discovered while dogfooding a local-mlx model end-to-end: onboarding, dev-server startup, env ordering, reasoning effort, and the system-prompt role.

## Changes

- Register Ollama (model list read live via `ollama list`) and a configurable local MLX/OpenAI-compatible server (`LOCAL_MLX_BASE_URL` / `LOCAL_MLX_MODEL_ID`) as providers in `packages/adapters/src/local-providers.ts`.
- Fix `pnpm dev`'s `WEB_PORT` env var not reaching the Vite child task under turbo's strict env mode.
- Fix onboarding's "connect a model" flow silently no-oping for keyless local providers: it gated on `apiKey` being non-empty, so leaving the (correctly optional) key field blank skipped `connect()` entirely and left the account on the `scripted`/OpenRouter fallback with no visible error.
- Fix local providers being registered before `.env` was loaded: `withLocalProviders()` ran at module-top-level (during ESM import evaluation), before the app's entrypoint called `loadRootEnv()`, so it always baked in the hardcoded fallback base URL/model instead of the configured one. Deferred to first use inside `run()`.
- Document `OLLAMA_BASE_URL` / `LOCAL_MLX_BASE_URL` / `LOCAL_MLX_MODEL_ID` in `.env.example`.
- Stop the local-mlx model reasoning at max effort on every turn: `mlx-openai-server`'s Qwen3.5 template defaults to `reasoning_effort: "xhigh"` unless told otherwise via `chat_template_kwargs`; now wired through `compat.thinkingFormat` so it respects Rakazo's own thinking-level policy.
- Send the system prompt as role `"system"` instead of pi's default `"developer"` role for local providers — `mlx-openai-server` rejects `"developer"` outright with a 422; reproduced against a live server with real message history and tools.
- Mock `withLocalProviders` in the tool-dispatch, computer, and attachment-forwarding runtime tests, whose bare `builtinModels()` stubs have no `setProvider` and would otherwise throw before the agent was ever constructed once every model lookup started routing through it.

## Test plan

- [x] `pnpm lint` — passes (fixed a biome import-order/formatting violation in `local-providers.ts` while packaging this PR)
- [x] `pnpm check` — 19/19 packages, typecheck clean
- [x] `pnpm test` — 580 passed, 49 skipped (fixed one test, `pi-runtime-attachments.test.ts`, that this change set broke — same missing-mock gap already fixed for the tool-dispatch/computer tests, just missed for this one)
- [x] `pnpm test:integration` — 44 passed
- [x] `pnpm test:e2e` — 15 passed

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,14 @@ SANDBOX_COMMAND_TIMEOUT_MS=300000
 OPENROUTER_API_KEY=
 PI_DEFAULT_PROVIDER=openrouter
 PI_DEFAULT_MODEL=deepseek/deepseek-v4-flash-0731
+# Local OpenAI-compatible providers registered by packages/adapters/src/local-providers.ts.
+# Both are optional — omit a var to skip registering that provider. If you also run Clawdia
+# locally, do NOT point LOCAL_MLX_BASE_URL at 127.0.0.1:8082 — that's Clawdia's own resident
+# mlx-openai-server (src/clawdia/mlx/daemon.py). Sharing it causes silent request contention
+# between the two apps. Give Rakazo its own dedicated mlx-openai-server on a different port.
+OLLAMA_BASE_URL=http://127.0.0.1:11434/v1
+LOCAL_MLX_BASE_URL=http://127.0.0.1:8081/v1
+LOCAL_MLX_MODEL_ID=DreamFoundries/Ornith-1.0-35B-6bit
 E2B_API_KEY=
 COMPOSIO_API_KEY=
 SUPERMEMORY_API_KEY=
--- a/apps/web/src/pages/Onboarding.tsx
+++ b/apps/web/src/pages/Onboarding.tsx
@@ -122,10 +122,19 @@ export function OnboardingPage() {
   async function saveModel() {
     setError(null);
     try {
-      if (apiKey) {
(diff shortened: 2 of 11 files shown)


bnsd55:
Thanks for pushing local models — the onboarding “keyless connect actually writes a credential” catch is real.

We don’t want to merge this as the product path because it adds **two branded providers** (Ollama via `ollama list`, and Local MLX hardcoded for `mlx-openai-server` on `:8081` with a default model id). That misses every other OpenAI-compatible server (Rapid-MLX on `:8000/v1`, LM Studio, llama.cpp, vLLM, Lemonade, a random `/v1` box).

We’re doing one generic provider instead: URL + model id + optional key, same wire for all of those.

- Issue (existing request): https://github.com/elie222/rakazo/issues/111
- PR: https://github.com/elie222/rakazo/pull/149

PR #105, opened 2026-08-21
feat(adapters): support local OpenAI-compatible model servers

izkold:
## What changed

Adds a `local` model provider so Rakazo can talk to an OpenAI-compatible model
server running on the operator's own machine — Ollama, LM Studio, llama.cpp, or MLX.

Pi's builtin catalog ships only hosted providers. A locally running model therefore
has no catalog entry, cannot be picked in onboarding, and cannot be set as the active
model. This registers one from environment configuration:

```bash
RAKAZO_LOCAL_MODELS=qwen3:4b,llama3.1:8b
RAKAZO_LOCAL_MODELS_URL=http://127.0.0.1:11434/v1   # optional, this is the default
RAKAZO_LOCAL_CONTEXT_WINDOW=32768                   # optional
RAKAZO_LOCAL_MAX_TOKENS=4096                        # optional
```

The provider is absent unless `RAKAZO_LOCAL_MODELS` names at least one model, so
deployments that don't use local models see no change at all. Models are reported at
zero cost, since they run on the operator's own hardware.

## Why

Rakazo advertises "bring your own model", and the self-host path is otherwise complete
without a cloud account. Local inference closes that gap: a self-hoster can run the
whole stack — Postgres, Docker computers, Graphile, and now the model — without a
hosted provider key or any outbound model traffic.

## Two details worth a reviewer's attention

**1. The keyless server still needs an `apiKey` resolver.**
`Models` treats a provider whose auth cannot resolve as unconfigured and hides its
models from the catalog. So `resolve` returns a placeholder string that
OpenAI-compatible local servers ignore. It is not a credential. If there's a preferred
way to express "this provider needs no auth", I'm happy to switch to it.

**2. `pi-runtime` now builds its catalog lazily instead of at module load.**
This one is a genuine trap. Entry points such as `apps/worker/src/index.ts` call
`loadRootEnv()` *after* their imports, but ESM hoists imports — so the adapters module
body executes before `.env` is loaded. A module-level env read there sees an empty
environment and silently registers nothing, surfacing at runtime as
`Unknown model local/<id>`. Building on first use fixes it. The previous
`const catalogModels = builtinModels()` was safe only because it read no env.

## Testing

- `pnpm lint` — clean (386 files)
- `pnpm check` — 19/19 tasks
- `pnpm test` — adapters suite 306 passed; new `pi-local-provider.test.ts` covers the
  unconfigured case, verbatim model ids, zero cost, base-URL default and override, and
  conditional registration
- Manual end-to-end against Ollama serving `qwen3:4b`: provider appears in the catalog,
  is selectable in onboarding, becomes the active model, and a bot replies

## Notes / possible follow-ups

- `reasoning` is hardcoded `false` for local models. That's wrong for reasoning models
  such as qwen3, which emit reasoning tokens before content. It's cosmetic in practice
  (the OpenAI-compatible path returns them in a separate field) but a per-model flag
  would be more accurate. Left out to keep this change small — happy to add it.
- Only `openai-completions` is wired up, which is what Ollama and LM Studio both serve.


izkold:
Thanks — good catch on the token limits.

Fixed in `deda472`: `RAKAZO_LOCAL_CONTEXT_WINDOW` and `RAKAZO_LOCAL_MAX_TOKENS` now go through a `tokenLimit` helper that requires a finite positive integer and throws otherwise. The old `Number(value) || default` would have accepted `-5` as a context window and quietly turned a typo into the default.

Added test coverage for the rejected inputs (`abc`, `-5`, `0`, `3.7`, `Infinity`, `1e400`) plus the defaults-and-valid-override path.

`pnpm lint`, `pnpm check` (19/19), and the adapters suite (308 passed) are all clean.

PR #154, opened 2026-08-24
fix(core): honor day-of-week and timezone in nextCronDate

julioblaq:
## Problem

`nextCronDate` only matched the cron **minute and hour** fields and scanned forward ~24 hours, so the day-of-week field was silently ignored and the `timezone` argument was discarded (`void timezone`):

```ts
const [minuteExpr, hourExpr] = parts;   // dow never read
...
if (matchField(minuteExpr...) && matchField(hourExpr...)) {
  void timezone;                         // accepted but unused
  return candidate;
}
```

Effect on routines with weekly/weekday schedules: they re-queue **daily**. Observed on a live instance:

- A `0 15 * * 5` routine (Friday 3 PM) that fired Friday 15:00 re-queued for **Saturday** 15:00.
- A `0 9 * * 1` routine (Monday 9 AM) that fired Monday 13:52 re-queued for **Tuesday** 13:00.
- A `0 7 * * 6` routine (Saturday 7 AM) re-queued for **Sunday**.

Additionally, cron hours were always evaluated as UTC, so a routine configured at 7:00 with timezone `America/New_York` fired at 3 AM Eastern.

## Fix

- Match the **day-of-week** field, mapping `7` to Sunday so both spellings work.
- Evaluate minute, hour, and weekday **in the routine's timezone** using `Intl.DateTimeFormat`, with a fallback to UTC for unknown identifiers.
- Extend the scan window from 24 hours to 8 days so any weekly slot is reachable (minute granularity retained; worst case ~11.5k formatter calls, well under 100 ms).

## Tests

Five new cases in `packages/core/src/cron.test.ts` covering: same-day daily slots, weekly day-of-week selection (the exact Monday→next-Monday case above), day 7 as Sunday, timezone-aware hour evaluation, and the UTC fallback. All 12 tests in the file and the full `@rakazo/core` suite pass locally.

Found while operating a self-hosted instance with six routines across five bots — happy to adjust anything.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduled task timing across time zones.
  * Added support for complex weekly schedules, including lists, ranges, stepped intervals, and Sunday aliases.
  * Invalid time zones now safely fall back to UTC.
  * Expanded scheduling calculations to reliably find the next occurrence over longer intervals.

* **Tests**
  * Added coverage for daily and weekly schedules, Sunday handling, time zones, and UTC fallback.

### DIFF
--- a/packages/core/src/cron.test.ts
+++ b/packages/core/src/cron.test.ts
@@ -4,6 +4,7 @@ import {
   cronFromPreset,
   describeCronPreset,
   formatCron,
+  nextCronDate,
   presetFromCron,
 } from "./cron.js";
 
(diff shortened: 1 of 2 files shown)


julioblaq:
Good catch on the composite day expressions — fixed in 7515430. Replaced the standalone-token rewrite with a dedicated `matchDayOfWeek` that treats 7 as Sunday inside plain values, comma lists, and ranges (with wraparound, so `5-7` = Fri/Sat/Sun). Added regression tests for `1,7` and `5-7`.

PR #207, opened 2026-08-25
Make production updater sidecar opt-in

elie222:
## Summary
- put the root-equivalent Docker updater behind an explicit Compose profile
- allow the default production stack to resolve without an updater token
- keep runtime token validation intact whenever the updater is enabled
- document and test both deployment modes

## Production failure addressed
The restricted production deploy account cannot edit `.env`. After the deploy-directory migration fixed the first interpolation failure, main reached the next new requirement and rolled back because the existing host has no `RAKAZO_UPDATER_TOKEN`. The normal stack does not use the updater, so it should not require or start the Docker-socket sidecar implicitly.

## Validation
- `POSTGRES_PASSWORD=fake-review-password docker compose --env-file /dev/null -f infra/compose/docker-compose.prod.yml config --services` excludes `updater` and needs no updater token
- the same command with `--profile updater` and a fake 32+ character token includes `updater`
- `pnpm test -- infra/updater/src/compose-service.test.ts infra/updater/src/updater-logic.test.ts` (28 passed)
- `pnpm lint`
- `git diff --check`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,8 +70,8 @@ RAKAZO_IMAGE_TAG=local
 RAKAZO_IMAGE_TAG_PREVIOUS=
 RAKAZO_UPDATER_IMAGE=ghcr.io/elie222/rakazo/updater
 RAKAZO_UPDATER_IMAGE_TAG=local
-# Required by infra/compose/docker-compose.prod.yml (sidecar refuses to start without it).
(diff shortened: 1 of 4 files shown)


elie222:
Addressed Greptile’s documentation finding in a923ad94: the first-deployment text now lists only the default-profile images and points to the opt-in updater command.

PR #208, opened 2026-08-25
Fix Codex stream reliability and surface provider failures

timvdhoorn:
## Summary

- Force SSE for OpenAI Codex OAuth models so mid-stream WebSocket closures cannot terminate runs with code 1006.
- Apply the same transport choice to nested agents.
- Propagate provider failures through the runtime iterator instead of emitting error text followed by a successful completion.
- Add regression coverage for transport selection and terminal error propagation.

## Verification

- Focused runtime tests: 9 passed.
- Adapter TypeScript check: passed after Prisma client generation.
- Biome: 4 files checked, no issues.
- Full adapter suite: 568 passed, 7 skipped, 3 unrelated failures. The unchanged upstream baseline reproduces the desktop sandbox inode-race failure and local shell startup stderr pollution in sandbox conformance tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider and agent failures now surface as actual errors instead of appearing as completed responses with error text.
  * Improved streaming reliability for OpenAI Codex Responses models by using SSE transport while preserving retry settings.

* **Tests**
  * Added regression coverage for error propagation, prompt failures, and streaming transport behavior.

### DIFF
--- a/packages/adapters/src/pi-runtime-error.test.ts
+++ b/packages/adapters/src/pi-runtime-error.test.ts
@@ -1,0 +1,62 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@earendil-works/pi-agent-core", () => ({
+  Agent: class {
+    state = { errorMessage: "WebSocket closed 1006", messages: [] };
+
+    subscribe() {}
+    async prompt() {}
+    async waitForIdle() {}
+    abort() {}
+  },
+}));
+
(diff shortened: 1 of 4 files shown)


timvdhoorn:
Addressed the CodeRabbit test-coverage note in eb5399d: Codex detection now has independent provider-only and API-only cases. Focused tests: 10 passed. Adapter TypeScript check and Biome both pass.

PR #218, opened 2026-08-25
feat(web): add a Board page — cross-bot Kanban of active work

deepanshu1422:
## Summary

Each bot's own chat only shows that bot's runs, so with several bots working
at once there's no single place to see what's queued, what's actively
running, what needs your input, and what's coming up next.

**Stacked on #217** (routine multi-schedule) — the upcoming-routine cards
show every schedule for a routine, not just one. This PR's branch is built
on top of that one, so the diff below includes #217's changes too until it
merges — once it does, GitHub will narrow this diff down to just the Board
page itself.

## Changes

- New `/app/board` route: three columns for active runs (Queued, Working,
  Needs you) plus an Upcoming column for routines sorted by next fire time
- Deliberately bounded to in-flight work only — no completed/failed history
  (that already lives in each bot's own chat), so the board can't grow
  unbounded
- Each card shows the bot, a routine indicator when relevant, and which
  computer (Team/Private) the bot is on; click through to that bot's thread
- Polls every 8s while the tab is visible, plus on focus

## Testing notes

- [x] `pnpm --filter @rakazo/web check`, `pnpm --filter @rakazo/api check`
- [x] `biome check` on touched files

## Summary by CodeRabbit

* **New Features**
  * Routines can now have multiple schedules, with support for editing, adding, and removing schedule entries.
  * Added a Board view showing queued, active, and waiting runs alongside upcoming routines.
  * Added Board navigation and authenticated access.
  * Routine-triggered runs now retain their originating routine information.
* **Improvements**
  * Schedule processing handles one-time, recurring, invalid, and empty schedules more reliably.
  * Routine lists display all configured schedules and offer a Stop action for running routines.


elie222:
Closing this for now. The Board view is extra feature noise — it's not clear what job it does that the sidebar / Now list doesn't already cover. We can revisit later if there's a clearer purpose.

PR #219, opened 2026-08-25
fix(web): keep progress chips visible for a minimum duration

deepanshu1422:
## Summary

A fast tool call (e.g. `schedule_create`) can complete before the browser
gets a paint frame, so the terminal event immediately strips the live
"working…" chip from the thread — the intermediate progress never actually
renders, even though the run did real work.

## Changes

- Track when a progress/steps message is first seen (`progressFirstSeenRef`)
- Delay applying a run's terminal event until that message has been visible
  for at least 700ms, but only if it ever actually rendered something
- The first-seen map is capped at 200 entries so a message whose terminal
  event never reaches this client (SSE drop mid-run, tab backgrounded
  through a reconnect, bot archived mid-run) doesn't linger forever

## Testing notes

- [x] `pnpm --filter @rakazo/web check`
- [x] `biome check` on the touched file

### DIFF
--- a/apps/web/src/lib/thread-events.test.ts
+++ b/apps/web/src/lib/thread-events.test.ts
@@ -11,11 +11,13 @@ import {
   computerPanelAutoBoot,
   computerTakeoverBlocked,
   isThreadSnapshotEvent,
+  MIN_PROGRESS_VISIBLE_MS,
   mergeThreadSnapshot,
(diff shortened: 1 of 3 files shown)


elie222:
Hey, why does the user need to see in the progress state if it's already done?

PR #220, opened 2026-08-25
feat: cross-bot @mention delegation

deepanshu1422:
## Summary

Typing `@BotB do X` in Bot A's 1:1 chat delegates the task to Bot B: it gets
its own Task/Run in its own thread, with the last 6 messages of Bot A's
conversation folded into the prompt as context. A live-updating
`delegated_task` card in Bot A's thread mirrors Bot B's status (queued →
running → completed/failed/cancelled).

## Changes

- `resolveDelegationTarget`: matches `@BotB` **anywhere** in the message
  (start, middle, or end — "can you ask @BotB to check this" works), not
  just as a prefix. Matched against every other bot's exact name (longest
  first, so "Chief of Staff" wins over "Chief"), as a whole word.
- `createDelegatedRun`: creates the Task/Run in Bot B's own thread inside
  the same transaction as the user's message, posts the `delegated_task`
  card in Bot A's thread
- `notifyDelegationOrigin`: mirrors Bot B's run status live into that card
  via the existing run.started/completed/failed hooks. The completion
  summary is tail-truncated rather than head-truncated — a verbose agent's
  actual conclusion is usually at the end of its response, not the start.
- New `runs.cancel` / `cancelRun`: cancels exactly one run, unlike the
  existing thread-wide `threads.stop` — needed so an accidental duplicate
  delegation (two Task/Run pairs queued on the same target bot) can be
  cleaned up without killing the run that's actually meant to keep going.
  The `delegated_task` card gets a hover Stop button wired to it.
- `Run.delegatedFromThreadId` / `delegatedFromMessageId`: plain columns, no
  FK (the origin thread/message can belong to a different bot than the run
  itself)

## Testing notes

- [x] `pnpm --filter @rakazo/api test` — new `resolveDelegationTarget`
      coverage (start/middle mention, longest-name-wins, bare mention with
      no task text, mentioned bot with no thread, no `@` at all)
- [x] `pnpm --filter @rakazo/adapters test`
- [x] `pnpm --filter @rakazo/api check`, `pnpm --filter @rakazo/web check`,
      `pnpm --filter @rakazo/adapters check`
- [x] `biome check` on touched files
- [x] Manually tested live against two real bots (Sarah/private) in a
      running instance — delegation triggers, the card mirrors status live,
      per-run cancel leaves the other bot's run untouched

### DIFF
--- a/apps/api/src/router.ts
+++ b/apps/api/src/router.ts
@@ -111,6 +111,7 @@ import { withSerializableRetry } from "./serializable-retry.js";
 import { assertTeachingSendAllowed, createTaughtSkillsService } from "./taught-skills.js";
 import { loadAllMessages, loadMessagePage } from "./thread-message-pages.js";
 import {
+  cancelRun,
   resolveThreadTarget,
   sendThreadMessage,
   setThreadUnreadState,
@@ -1001,6 +1002,12 @@ export function createRouter(deps: RouterDeps) {
         const target = await resolveThreadTarget(deps.prisma, context.actor, input);
         await setThreadUnreadState(deps.prisma, context.actor, target, true);
         return { ok: true as const };
+      }),
+    },
+    runs: {
+      cancel: authed.runs.cancel.handler(async ({ context, input }) => {
+        const result = await cancelRun(deps, context.actor, input.runId);
+        return result;
       }),
     },
     computer: {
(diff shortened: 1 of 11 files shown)


elie222:
Closing as a duplicate of #186, which is the 1:1 bot-to-bot path we're taking. Group handoff stays on main as handoff_to_bot.

PR #94, opened 2026-08-20
Reconcile stale pending Composio connections on catalog fetch

taur-us:
## Summary

A Composio connection can finish on Composio's side after our own completion check gave up (see `PluginsOverlay`'s polling timeout), leaving the local `Connection` row stuck at status `pending` even though Composio reports it connected. Callers that trust the local status — like the run executor — never learn the connection actually succeeded.

This reconciles stale `pending` rows whenever the plugin catalog is fetched: `connections.catalog` now diffs Composio's reported `connected` slugs against locally `pending` rows for that workspace/user and flips them to `connected` before returning the catalog to the caller.

## Test plan

- [x] `pnpm lint` — passes
- [x] `pnpm check` — passes (all 19 packages typecheck)
- [x] `pnpm test` — passes (578 passed, 49 skipped)
- [x] `pnpm test:integration` — passes (44 passed, needs Docker)
- [ ] `pnpm test:e2e` — blocked in my environment: Playwright's dev server port (5174) was already in use by an unrelated local process, unrelated to this change. Not run.

PR #232, opened 2026-08-26
Reconcile stale pending Composio connections and keep catalog on write failure

taur-us:
## Summary
- Reconcile stale pending Composio connections when the plugin catalog is fetched, so a half-finished connect attempt doesn't linger forever
- Don't blank the plugin catalog when the reconciliation write fails — serve the last good catalog instead

## Test plan
- pnpm vitest run apps/api (22 passed)
- tsc + biome clean for apps/api

## Summary by CodeRabbit

- **Bug Fixes**
  - Connection records are now automatically marked as connected when matching providers are detected in the catalog.
  - Catalog results remain available even if connection status updates cannot be completed.
  - Provider catalog errors continue to return an empty list.


elie222:
Already on main — same reconcile landed in #94 / #144 (pending rows flip to connected on catalog, write failure does not blank the list).

PR #236, opened 2026-08-26
Fix write_file silently corrupting JSON object content

taur-us:
## Summary
- Models sometimes pass `content` as a JSON object where a string is expected; `String()` coerced it to the literal text `[object Object]`, so the file was written corrupted while the tool still reported success
- Serialise object arguments to readable JSON instead, at both the pi-agent argument-normalisation layer and the executor's write_file handler

## Test plan
- New regression test in pi-runtime-tool-dispatch.test.ts asserting object content is serialised (fails with "[object Object]" before the fix)
- pnpm vitest run packages/adapters (302 passed); tsc + biome clean

### DIFF
--- a/packages/adapters/src/executor.ts
+++ b/packages/adapters/src/executor.ts
@@ -38,6 +38,7 @@ import {
   type ThreadEvents,
 } from "@rakazo/db";
 import { builtinAgentTools } from "./builtin-tools.js";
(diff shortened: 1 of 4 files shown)


elie222:
Superseded by #237 — same fix, onto current main.

PR #238, opened 2026-08-26
Default new routines to the browser's local timezone

taur-us:
## Summary
- The routine editor hardcoded `timezone: "UTC"` on create, so a "9:00 AM every day" schedule fired at 9:00 UTC regardless of where the user lives (hit in practice: a 9am news routine firing at 7pm Sydney time)
- New routines now send the browser's IANA timezone (`Intl.DateTimeFormat().resolvedOptions().timeZone`, falling back to `UTC` when unavailable) — the scheduler already handles IANA zones and DST correctly
- The editor shows which timezone the schedule times refer to (the routine's stored zone when editing, the detected local zone when creating)

Deliberately left out: mobile create path and chat-native /schedule tools (bots can pass an explicit timezone); a workspace-level default timezone feels like a separate design discussion.

## Test plan
- New unit test for the helper; pnpm vitest run apps/web (102 passed)
- Existing routine e2e seeds via RPC with explicit timezones and asserts timezone preservation on edit — unaffected
- tsc + biome clean for apps/web

## Summary by CodeRabbit

* **New Features**
  * Routine schedules now automatically use the browser’s local timezone.
  * The routine editor displays the timezone applied to schedule times.
  * New routines preserve the detected local timezone, with UTC used as a fallback.

* **Bug Fixes**
  * Improved timezone handling for routine scheduling across different regions.
  * Schedule times now remain consistent with the timezone detected by the browser.

### DIFF
--- a/apps/web/src/lib/local-timezone.test.ts
+++ b/apps/web/src/lib/local-timezone.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from "vitest";
+
+import { localTimezone } from "./local-timezone.js";
+
+describe("localTimezone", () => {
+  it("returns the browser's IANA timezone", () => {
+    const zone = localTimezone();
+    expect(zone).not.toBe("");
+    // IANA zones contain a slash; the UTC fallback does not.
+    if (Intl.DateTimeFormat().resolvedOptions().timeZone) {
+      expect(zone).toContain("/");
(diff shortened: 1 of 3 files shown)


elie222:
Superseded by #239 — same change, with the unit test fixed so UTC CI runners pass.

PR #249, opened 2026-08-26
feat(web): add Korean localization

inhodev:
## Summary

- add typed Korean UI copy with an English fallback
- localize onboarding, chat, settings, models, voice, integrations, MCP, and runtime activity
- allow self-hosted deployments to set a default UI locale with `VITE_DEFAULT_UI_LOCALE`
- apply the deployment default at container startup so pulled prebuilt images honor it too

## Localization approach

- preserve product and protocol names such as Rakazo, Bot, OpenAI, Codex, OAuth, MCP, and SuperGrok
- prefer natural, user-facing Korean over literal technical translations
- resolve locale from saved preference, deployment default, browser language, then English fallback

## Review follow-up

- narrowed activity-prefix translation so ordinary model prose is preserved
- completed the flagged UI copy and accessible labels
- corrected Korean routine word order
- localized approval-rule targets without losing category/connector/tool context

## Verification

- `pnpm lint`
- `pnpm --filter @rakazo/web check`
- `pnpm --filter @rakazo/web test` — 21 files, 114 tests passed
- production web build
- runtime locale file served from a production preview with `ko-KR`
- Compose YAML syntax parsed for both development and production files

## Scope

This PR contains only the Korean localization work and its deployment configuration. Follow-up mobile drawer and active-run timer fixes remain excluded.


elie222:
Superseded by #250 — same Korean localization, rebased onto current main with conflict resolution. Credit @inhodev (kiminho).

PR #348, opened 2026-08-28
fix(db): make group index migrations retry-safe

luinbytes:
Follow-up to #345 after CodeRabbit identified a deployment retry edge case after the feature PR had already merged.

- tolerate indexes already created by a partially completed migration
- tolerate the legacy index already being absent
- assert archive run discovery remains scoped to the group thread

Verification:
- `npm exec -- biome check packages/db/src/groups.test.ts`
- `npm exec -- vitest run packages/db/src/groups.test.ts` (2 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability by safely creating indexes only when needed.
  * Repaired invalid database indexes during migration.
  * Prevented migration failures when attempting to remove an index that is already absent.
  * Ensured archiving a group only cancels runs associated with that group’s thread.

* **Tests**
  * Added coverage verifying thread-specific run cancellation during group archiving.


luinbytes:
@coderabbitai review

PR #364, opened 2026-08-28
fix(scripts): pass the root .env to Compose in backup and restore

Scottcjn:
Since #343 the Compose file requires `SANDBOX_SUPERVISOR_TOKEN` and `SCREEN_PROXY_SECRET` via `${VAR:?}` interpolation. Compose resolves its default env file relative to the Compose file (`infra/compose/`), not the repo root, so `./scripts/backup.sh` and `./scripts/restore.sh` now abort before doing anything:

```
error while interpolating services.web.environment.SCREEN_PROXY_SECRET:
required variable SCREEN_PROXY_SECRET is missing a value
```

Every other Compose invocation in the repo (`pnpm compose:up`, the self-host docs, `backup-prod.sh`) already passes `--env-file .env`. This does the same in the two scripts, using the repo-root `.env` the docs tell operators to create.

**Tested:** reproduced the failure on `main`, then ran `./scripts/backup.sh` and `./scripts/restore.sh` end-to-end against `docker compose up -d postgres` from a fresh checkout with `.env` copied from `.env.example` (dump written, psql restore succeeded). `bash -n` and `shellcheck` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved backup and restore reliability by consistently loading environment settings from the project configuration.
  - Added support for environment variables when the project configuration file is unavailable.
  - Standardized service startup, database readiness checks, database restoration, and backup steps to use the configured Compose environment.
  - Existing backup directory handling remains unchanged.


Scottcjn:
Updated: `--env-file` is now added only when the root `.env` exists, so setups that export the variables in the environment (cron/CI) keep working exactly as before.

PR #365, opened 2026-08-28
fix(sandbox): fail Docker stop and destroy on a supervisor error

Scottcjn:
`DockerSandboxProvider.stop` and `destroy` discarded the supervisor's response, so a 401 from a rotated `SANDBOX_SUPERVISOR_TOKEN`, a 500, or any other failure resolved as success.

Downstream effects:
- `replaceComputer` then nulled `providerRef` and left the `rakazo-bot` container running with nothing referencing it.
- `sleepComputerIfIdle` marked the computer suspended while it was still up; its revert-to-running `catch` could never fire.

This checks the status the same way `releaseScreen` in the same file already does: 404 means the supervisor no longer has the container (the state we want), anything else throws so the caller sees it. Two unit tests added.

**Tested:** `docker-sandbox`, `computer-idle`, `computer-lifecycle` tests pass; `pnpm lint` clean; `@rakazo/adapters` tsc clean; full `pnpm test` green.

Happy to make 404 throw as well if you'd rather stop/destroy be strict — I matched `releaseScreen`'s tolerance.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox shutdown reliability by reporting failures when stopping or removing a sandbox does not succeed.
  - Treats already-removed sandboxes as successfully stopped or destroyed, avoiding unnecessary errors.
  - Includes relevant server error details when cleanup operations fail.
  - Added coverage for error handling during sandbox cleanup.


Scottcjn:
Follow-up after an extra review pass on our side: the error now includes the supervisor's response body (first 200 chars), and the test covers 500 as well as 401.

One thing worth being explicit about since it's a deliberate contract change: the two router paths that `await sandbox.stop(...)` before writing `state: "stopped"` will now surface a supervisor failure instead of completing the switch with an orphaned container. Retries are safe — if the container did stop despite a non-2xx, the retry's `stop` gets a 404 and proceeds.

PR #366, opened 2026-08-28
fix(supervisor): return container stderr instead of folding it into stdout

Scottcjn:
The supervisor runs exec without a TTY, so Docker multiplexes the stream and tags every frame as stdout or stderr. `stripDockerStream` dropped the tag and concatenated everything, and `runContainerCommand` then reported `stderr: ""` unless the command timed out.

Three effects:
1. Every `result.stderr || "..."` diagnostic in the supervisor could only ever show the generic fallback.
2. Docker computers never emitted a stderr event to the agent.
3. Anything a process wrote to stderr was spliced into stdout — which the file read (base64) and list (JSON) endpoints parse verbatim.

This adds `demuxDockerStream` in `supervisor-logic.ts` (frames split by type byte; payloads concatenated as buffers before decoding so multibyte UTF-8 across frame boundaries stays intact) and returns both streams. On timeout the timeout line is appended to the real stderr. The file-write error path prefers stderr and falls back to stdout. Tests added.

**Tested:** supervisor tests 52/52, supervisor tsc clean, `pnpm lint` clean, full `pnpm test` green.

Note this changes observable output — agents will now see stderr as a separate event rather than inline in stdout — so worth a look at whether any prompt relied on the old behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container command output handling by separating standard output from error output.
  * Error messages from failed file writes are now reported more accurately.
  * Timeout-related error details are preserved in command results.
  * Added support for correctly handling both multiplexed Docker output and raw terminal streams.


Scottcjn:
Extra review pass on our side turned up two hardening items, pushed:

- `demuxDockerStream` now keeps the bytes that actually arrived when the stream is cut mid-frame instead of skipping past the buffer end (test added, including a dangling <8-byte header).
- The timeout line no longer glues onto the last stderr line when the command's stderr lacks a trailing newline.

Also for the record: the raw-vs-multiplexed heuristic (first byte ≤ 2) is unchanged from the removed `stripDockerStream`, so raw streams behave exactly as before.

Scottcjn:
Verified the validate-before-demux commit locally — 54/54 supervisor tests, tsc and biome clean. The stricter whole-buffer validation is a better trade than my partial-frame salvage: it removes the false-positive on raw streams that happen to start with 0x00–0x02, and a genuinely cut stream degrades to raw stdout instead of guessing. LGTM from our side.

PR #376, opened 2026-08-29
fix(computer): route Docker browsers through Rakazo wrapper

### LINKED ISSUE #375: [Bug]: open_path and launch_app fail: xdg default browser handler bypasses rakazo-browser wrapper
### Confirmation

- [x] This is not a security vulnerability report

### Description

open_path and launch_app fail inside the Docker computer sandbox. The xdg default browser handler points at bare /usr/bin/chromium, which cannot start in the container without --no-sandbox. The rakazo-browser wrapper that passes the correct flags is installed at /usr/local/bin/ but is never registered as a .desktop handler, so anything routed through xdg crashes. Agents effectively lose two of the four graphical tools, which makes an image packaging issue look like a model failure.

### Steps to reproduce

Self-host with SANDBOX_PROVIDER=docker via infra/compose/docker-compose.yml
Build the computer image: docker build -t rakazo/computer:local infra/sandboxes/computer
Create a bot with a vision-capable model and send it any task requiring a browser
Observe open_path and launch_app both returning 500
Exec into the bot container and confirm the cause:

```
$ which rakazo-browser xdg-open chromium
/usr/local/bin/rakazo-browser
/usr/bin/xdg-open
/usr/bin/chromium

$ xdg-settings get default-web-browser
chromium.desktop

$ grep Exec /usr/share/applications/chromium.desktop
Exec=/usr/bin/chromium %U

$ ls /usr/share/applications/ | grep -i rakazo
(no output)

$ DISPLAY=:1 xdg-open https://example.com
[ERROR:zygote_host_impl_linux.cc:128] No usable sandbox! ...
```

### Expected behavior

open_path and launch_app open URLs and applications through the rakazo-browser wrapper, using the same flags and persistent profile as the XFCE desktop menu entry, and succeed.

### Actual behavior

open_path returns 500 for any URL. launch_app can technically open a working browser, but only if the caller passes application="rakazo-browser" exactly, an internal implementation name that appears nowhere in the tool's description, in the system prompt, or anywhere else visible to the model. 
Nothing tells an agent this name exists or that it differs from the browser names it would naturally try (chrome, chromium, firefox), all of which fail. 
From the model's perspective, both tools are simply broken; the working path only exists if a human has told it in advance which magic string to use. This defeats the purpose of the tool: an agent shouldn't need out-of-band, environment-specific knowledge to open a browser.

### Rakazo version or commit

86405ac

### Operating system

Linux

### SANDBOX_PROVIDER

_No response_

### AGENT_RUNTIME

_No response_

### Logs or screenshots

```shell

```

bferanmi806-sketch:
Fixes #375

Docker browser actions could bypass `rakazo-browser` in two ways: `xdg-open` resolved to Chromium's desktop handler, and `launch_app` accepted browser aliases that executed the raw binaries directly. In the container that drops the required browser flags and profile handling.

Register `rakazo-browser` as the Docker image's HTTP/HTTPS/HTML XDG handler and canonicalize browser launch aliases through the same wrapper for both the computer-control path and Docker-exec fallback.

The wrapper also selects the existing per-display Chromium profile for Team Computer screens while preserving an explicit `--user-data-dir`. Non-browser launches such as `xterm` are unchanged.

Verification:
- focused tests: 82 passed, 1 skipped on Windows
- supervisor typecheck passed
- changed-path Biome check passed
- shell/Python syntax checks passed
- `git diff --check` passed
- `pnpm check` passed

Full `pnpm test` and repository-wide lint still hit pre-existing Windows/environment failures. A Docker image build was attempted, but Docker Desktop was not running in the local environment.

## Summary by CodeRabbit

* **New Features**
  * Added Rakazo Browser as the default handler for web links and HTML files.
  * Added isolated browser profiles for different displays.
  * Browser launch aliases now route through the supported Rakazo Browser launcher.

* **Bug Fixes**
  * Prevented direct launches of unsupported browsers.
  * Preserved explicitly provided browser profile settings.
  * Startup now reports errors if default browser registration fails.

* **Tests**
  * Added coverage for browser routing, profile isolation, default handlers, and batch actions.

### DIFF
--- a/infra/sandboxes/computer/Dockerfile
+++ b/infra/sandboxes/computer/Dockerfile
@@ -33,6 +33,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     fonts-liberation \
     fonts-dejavu-core \
     dbus-x11 \
+    xdg-utils \
   && rm -rf /var/lib/apt/lists/* \
   && groupadd --gid 1000 rakazo \
   && useradd --uid 1000 --gid rakazo --create-home --shell /bin/bash rakazo \
@@ -49,6 +50,7 @@ COPY --chmod=755 start.sh /usr/local/bin/rakazo-computer
 COPY --chmod=755 control.py /usr/local/bin/rakazo-computer-control
 COPY --from=capture-builder /build/librakazo-xcapture.so /usr/local/lib/librakazo-xcapture.so
 COPY --chmod=755 rakazo-browser /usr/local/bin/rakazo-browser
+COPY --chmod=644 rakazo-browser.desktop /usr/share/applications/rakazo-browser.desktop
 COPY --chmod=644 fluxbox.init /etc/rakazo/fluxbox/init
 COPY --chmod=644 fluxbox.apps /etc/rakazo/fluxbox/apps
 COPY --chmod=644 fluxbox.menu /etc/rakazo/fluxbox/menu
--- a/infra/sandboxes/computer/control.py
+++ b/infra/sandboxes/computer/control.py
@@ -18,11 +18,6 @@ MAX_ARGV = 32
 MAX_ARG_LEN = 16_384
 KNOWN_LAUNCH = frozenset(
     {
-        "chromium",
-        "chromium-browser",
-        "firefox",
-        "google-chrome",
-        "google-chrome-stable",
         "rakazo-browser",
         "xterm",
     }
--- a/infra/sandboxes/computer/rakazo-browser
+++ b/infra/sandboxes/computer/rakazo-browser
@@ -1,23 +1,24 @@
 #!/bin/sh
 set -eu
 RAKAZO_HOME="${HOME:-/home/rakazo}"
-PROFILE="$RAKAZO_HOME/.browser-profiles/chromium"
+case "${DISPLAY:-:1}" in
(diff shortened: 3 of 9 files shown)


cyrilS-dev:
Thanks for the fast turnaround, confirmed fixed on my end (xdg-open now routes through rakazo-browser correctly).

 I'd also hit a separate stale Chromium SingletonLock issue on shared Team Computers, where a killed container left a lock file behind and the next container's Chromium refused to start ("profile appears to be in use by another Chromium process").
 Wondering if the per-display profile isolation in this PR might fix that too, will test and report back.

bferanmi806-sketch:
Glad that fixed it! If the SingletonLock issue still shows up after you test the per-display profiles, feel free to tag me. Happy to take a look at that too.

cyrilS-dev:
thanks @bferanmi806-sketch 
Following up on this after pulling and rebuilding, I'm now seeing what looks like a regression on the Computer: the screen renders mostly gray in both the Rakazo UI and raw noVNC, and only the specific regions I click or hover on redraw with actual content. This wasn't happening before this merge, same setup otherwise. Looks like an XDamage-related capture issue (x11vnc runs without -noxdamage in start.sh, so it only pushes incrementally-damaged regions, and something about the startup sequence now seems to miss the initial full-frame damage event).

image

It's better to open a separate issue for this?

bferanmi806-sketch:
Yeah, I think a separate issue makes sense since this looks distinct from the original routing problem. Tag me there and I’ll take a look.

PR #391, opened 2026-08-30
fix(adapters): expose schedule tools in group chats

luinbytes:
## Summary

- expose schedule_create, schedule_list, and schedule_cancel in group-chat tool registries
- persist the originating thread for tool-created routines so group schedules wake back into the group
- re-authorize the stored target at wake time and fall back to the bot DM if the group is gone, archived, or no longer contains the bot
- preserve existing behavior for old and UI-created routines without a stored thread

## Root cause

The shared built-in tool filter explicitly removed every schedule tool from group threads. The schedule persistence path also discarded the originating thread, so simply removing that filter would have sent later wakeups to the bot DM.

## Verification

- focused schedule/executor tests: 28 passed
- @rakazo/db and @rakazo/adapters TypeScript checks: passed
- Prisma schema validation: passed
- Biome on changed TypeScript files: passed

## Migration safety

The nullable thread target keeps existing rows compatible. Its foreign key is added NOT VALID and validated in a separate migration to avoid holding the stronger add-constraint lock during the validation scan.

## Summary by CodeRabbit

* **New Features**
  * Scheduling tools now work in group conversations.
  * Routines can target specific conversation threads.
  * Routine wakeups, tasks, and notifications use the configured thread when valid, with a fallback to the default thread.
  * Schedule lists and cancellations are scoped to the active conversation when applicable.

* **Bug Fixes**
  * Improved handling when a target thread is unavailable or inactive.
  * Routine data remains associated with its conversation thread.

### DIFF
--- a/packages/adapters/src/executor.test.ts
+++ b/packages/adapters/src/executor.test.ts
@@ -10,6 +10,8 @@ describe("createRunExecutor", () => {
     const cancel = vi.fn(async () => undefined);
     const append = vi.fn(async () => undefined);
     const updateMany = vi.fn(async () => ({ count: 1 }));
+    const taskCreate = vi.fn(async () => ({ id: "task-1" }));
+    const runCreate = vi.fn(async () => ({ id: "run-1" }));
     const prisma = {
       routine: {
         findUnique: vi.fn(async () => ({
@@ -22,6 +24,7 @@ describe("createRunExecutor", () => {
           timezone: "UTC",
           active: true,
           nextRunAt: scheduledAt,
+          threadId: "group-thread-1",
         })),
       },
       bot: {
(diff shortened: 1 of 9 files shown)


luinbytes:
Addressed Greptile’s destination finding in 35b3d97: tool-created routines now retain the originating thread, wakeups re-authorize that DM/group target before use, and the executor regression test proves the task, run, and fired event stay on the group thread. Existing routines keep the DM fallback. The nullable foreign key is added and validated in separate migrations.

Fresh verification: 28 focused tests passed; adapters/db typechecks, Prisma validation, and Biome passed.

luinbytes:
Addressed Greptile’s cross-thread finding in a930e75. Group schedule_list and schedule_cancel now include the current group thread in their database scope; DM calls deliberately omit that filter so legacy and UI-created bot routines remain visible and manageable there. The focused test covers both the narrowed group queries and unchanged DM queries.

Fresh verification: 28 focused tests passed; adapters typecheck and Biome passed.

PR #397, opened 2026-08-30
feat(web): show chat message timestamps on hover

davidrobertson:
## Summary

- show each chat message's real creation time beside the existing reply/copy pill
- reveal the timestamp on message hover and when the action controls receive keyboard focus
- keep the metadata subdued and semantic with a `<time>` element

## Verification

- `NODE_ENV=test pnpm exec vitest run apps/web/src/components/MessageHoverMetadata.test.tsx` - 1 passed
- `pnpm --filter @rakazo/web check` - passed
- isolated Playwright render at the committed revision confirmed:
  - `6:14 PM` formatting from the real `createdAt`
  - opacity 0 initially, 1 on hover, and 1 on keyboard focus
  - 6px spacing between timestamp and action pill
  - no clipping or overlap in the isolated fixture

The regression test was first run before implementation and failed because the expected timestamp was absent.

## Residual verification

The local API was unavailable during visual QA, so the render used an isolated message fixture rather than an authenticated live transcript. A fresh parent Biome invocation was also terminated by the local host without returning diagnostics; repository CI remains the authoritative lint gate.

## Summary by CodeRabbit

* **New Features**
  * Message timestamps now appear alongside Reply and Copy actions when hovering over or focusing a message.
  * Metadata visibility and positioning are presented consistently around message actions.
  * Timestamps follow the active language and regional formatting settings.

* **Bug Fixes**
  * Corrected localized time display while retaining the underlying date and time information.

* **Tests**
  * Added coverage for English and German timestamp formatting, metadata ordering, and hover/focus visibility behavior.

### DIFF
--- a/apps/web/src/components/MessageHoverMetadata.test.tsx
+++ b/apps/web/src/components/MessageHoverMetadata.test.tsx
@@ -1,9 +1,18 @@
+import { i18n } from "@lingui/core";
+import { afterEach, describe, expect, it } from "vitest";
 import { renderToStaticMarkup } from "react-dom/server";
-import { describe, expect, it } from "vitest";
 import { MessageHoverMetadata } from "./MessageHoverMetadata";
 
 describe("MessageHoverMetadata", () => {
+  afterEach(() => {
+    i18n.load("en", {});
+    i18n.activate("en");
+  });
+
   it("shows the message's creation time beside its actions on hover or focus", () => {
+    i18n.load("en", {});
+    i18n.activate("en");
+
     const createdAt = new Date(2026, 7, 21, 18, 14).toISOString();
     const html = renderToStaticMarkup(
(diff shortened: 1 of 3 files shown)


davidrobertson:
CI is blocked only by the Postgres journeys flake at `packages/testkit/src/journeys.test.ts:1575`:

```text
AssertionError: expected <ask message id> to be <new ask message id>
1 failed, 64 passed
```

This is the same message-ID race that failed on #395 and passed on its replacement run. The timestamp diff does not touch the testkit, API, database, or message lifecycle. All changed-area checks, builds, Web E2E, lint, typecheck, unit tests, CodeRabbit, and Greptile pass.

I attempted `gh run rerun 33314256603 --failed`, but GitHub rejects reruns from the fork author without upstream admin rights. Please rerun the failed Postgres journeys job.

PR #427, opened 2026-08-31
feat: add OrcaRouter as a first-class model provider

zsanig22-dotcom:
Rakazo users who bring their own model credentials through Pi can now select **OrcaRouter** as a first-class provider and paste an `ORCAROUTER_API_KEY`, the same way they already do for OpenRouter — instead of fitting the gateway behind a custom OpenAI-compatible base URL. That single key turns on OrcaRouter's adaptive routing, automatic failover, observability, and guardrails across the same model namespace Pi already speaks (`provider/model` ids like `orcarouter/auto`), so persistent AI teammates get the whole stack without any app-level code changes.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a named provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The change mirrors Rakazo's existing OpenRouter wiring exactly: a `pi-orca-router-provider.ts` registers a fixed-base-URL provider on the Pi catalog, `deployment-model.ts` pairs `ORCAROUTER_API_KEY` with the provider so a deployment default never hands one vendor's key to another, the runtime falls back to the OrcaRouter env key for `orcarouter` runs, and `model-vision.ts` / `auto-review.ts` gate vision and safety-check the same way they do for OpenRouter. `PI_DEFAULT_PROVIDER=orcarouter` also synthesizes a selectable entry when `PI_DEFAULT_MODEL` names a model newer than the static catalog.

Verified: `pnpm test` (2002 passing), `pnpm check`, and `pnpm lint` are green, and the new live canary in `packages/testkit/src/providers.canary.test.ts` streams a reply from `orcarouter/auto` through the real gateway. Docs updated in `.env.example`, `README.md`, `docs/self-host.md`, and `SETUP_PROMPT.md`.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

## Summary by CodeRabbit

* **New Features**
  * Added Orca Router as a supported model provider.
  * Added support for configuring Orca Router with `ORCAROUTER_API_KEY`.
  * Added access to Orca Router’s reasoning and vision-capable models, including the `orcarouter/auto` default.
  * Added provider support to deployment configuration, model selection, runtime features, and live connectivity checks.

* **Documentation**
  * Updated local setup and self-hosting instructions to include Orca Router configuration.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,7 @@ SANDBOX_COMMAND_TIMEOUT_MS=300000
 MAX_TOOL_CALLS_PER_TURN=
 # Deployment-wide fallback model. PI_DEFAULT_PROVIDER picks which key below is used.
 OPENROUTER_API_KEY=
+ORCAROUTER_API_KEY=
 ANTHROPIC_API_KEY=
 PI_DEFAULT_PROVIDER=openrouter
 # Text-only by default. Computer use (screenshots via computer_observe / computer_act)
--- a/README.md
+++ b/README.md
@@ -73,7 +73,8 @@ cp .env.example .env
 
 Set `BETTER_AUTH_SECRET`, `ENCRYPTION_KEY`, and `SCREEN_PROXY_SECRET` in `.env` to independent
 long random values. Docker sandboxes also need a dedicated `SANDBOX_SUPERVISOR_TOKEN`. You can
-also set `OPENROUTER_API_KEY`, or connect a supported model provider during onboarding.
+also set `OPENROUTER_API_KEY` or `ORCAROUTER_API_KEY`, or connect a supported model provider
+during onboarding.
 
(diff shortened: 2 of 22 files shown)


elie222:
Skipping this as a forked provider. Use the existing OpenAI-compatible slot instead of adding OrcaRouter.

PR #496, opened 2026-09-02
fix(compose): warn on arm64 hosts using edge image tags

fetw882:
## Summary
- On `arm64`/`aarch64`, warn if `RAKAZO_IMAGE_TAG` or `RAKAZO_COMPUTER_IMAGE_TAG` is still `edge` (amd64-only), pointing at `docs/self-host.md`
- Script-only; no CDN defaults

**Note:** Branch is rebased onto #495 (`fix/installer-compose-pull-retry`) so both installer knives share one linear history. Prefer merge **#495 first**, then this PR (or merge this after #495 lands on main and we rebase). Diff vs main currently includes #495 pull-retry commits until that lands.

## Test plan
- [ ] `bash -n infra/compose/install-images.sh`
- [ ] arm64 + edge tags → stderr warning
- [ ] arm64 + pinned tags → silent
- [ ] amd64 host → silent

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ARM64 image compatibility warnings by accurately detecting effective Docker Compose image tags.
  * Shell-exported image-tag values now take precedence when Compose environment resolution is unavailable, including intentionally empty values.
  * Improved fallback handling for `.env` values, quoted or commented entries, whitespace, and chained or concatenated default expressions.
  * Preserved image-pull retry behavior, with up to three attempts and a two-second delay between attempts.

* **Tests**
  * Added coverage for environment precedence, empty values, and fallback expansion scenarios.

### DIFF
--- a/infra/compose/install-images-fallback.smoke.sh
+++ b/infra/compose/install-images-fallback.smoke.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# Offline smoke for the arm64-warn fallback in install-images.sh:
+# unset vs empty shell, .env, and ${NAME:-def} / ${NAME-def}.
+set -euo pipefail
+
+root="$(cd "$(dirname "$0")" && pwd)"
+src="$root/install-images.sh"
+
+extract_fn() {
+  local name="$1"
+  awk -v s="$name" '
+    $0 ~ "^    " s "\\(\\) \\{" { p = 1 }
+    p { print substr($0, 5) }
+    p && $0 ~ /^    }$/ { exit }
+  ' "$src"
+}
+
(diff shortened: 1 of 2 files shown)


fetw882:
Addressed Greptile's resolve_env_default gap on the else fallback path: `${BASE_TAG:-edge}`-style refs are now expanded (shell still wins over `.env`). Smoke: unset → edge; BASE_TAG=v1.2.3 → v1.2.3. Commit 84f17d3.

fetw882:
Addressed Greptile P1 in `66e8711`: fallback `resolve_env_default` now expands concatenated interpolations (e.g. `${PREFIX}edge`) consistent with Compose; added `lookup_env_value`; smoke passed.

PR #231, opened 2026-08-26
Add Ollama and local MLX server as OpenAI-compatible model providers

taur-us:
## Summary
- Register Ollama (`ollama list` discovery) and a self-hosted mlx-openai-server as OpenAI-compatible providers, keyless and opt-in via env vars, so bots can run fully local models
- Fix onboarding silently failing to set a keyless local model as default; document provider env vars in .env.example
- Local-provider fixes found while testing: register after .env load, stop local-mlx reasoning defaulting to max effort, send system prompt as role "system" (mlx-openai-server rejects the developer role), mock `withLocalProviders` in tests that stub the model catalog
- Wire `model.reasoning` through to PiAgentRuntime's thinkingLevel

## Test plan
- pnpm vitest run packages/adapters packages/contracts (310 passed)
- tsc clean for adapters, contracts, apps/web; biome clean

## Summary by CodeRabbit

* **New Features**
  * Added support for local Ollama and MLX model providers.
  * Local models can be discovered automatically, configured with custom endpoints, and used without API keys or usage costs.
  * Added local-provider details to the model catalog, including keyless access and local execution indicators.
  * Reasoning settings now adapt automatically to the selected model’s capabilities.

* **Bug Fixes**
  * Keyless local providers can now be connected during onboarding without an API key.

* **Tests**
  * Added coverage for local provider discovery, reasoning behavior, and compatibility with local model servers.

PR #514, opened 2026-09-03
feat(adapters): add local-mlx provider for self-hosted MLX OpenAI-compatible servers

taur-us:
## Why

Rakazo's generic `local` provider (`RAKAZO_LOCAL_MODELS` / `RAKAZO_LOCAL_MODELS_URL`) and the per-user `openai-compatible` connect both target OpenAI-compatible servers, but neither encodes the compatibility settings MLX servers actually require:

- `mlx-openai-server` (and Rapid-MLX) rejects pi's default `"developer"` system role with a `422`; it only accepts `"system"`.
- Qwen chat templates read `chat_template_kwargs.enable_thinking`. Without a thinking-format hint, those servers default to `reasoning_effort: "xhigh"` and can burn the entire `maxTokens` budget on thinking before reaching an answer.

The generic `local` provider sets `reasoning: false` and no `compat`; the `openai-compatible` provider sets `supportsDeveloperRole: false` but disables reasoning and has no thinking-format option. Pointing either at an MLX model misbehaves out of the box. This adds a dedicated, deployment-wide `local-mlx` provider so an MLX endpoint and an Ollama endpoint can coexist, each with the right settings.

## What changed

- `packages/adapters/src/pi-local-provider.ts`: new `local-mlx` provider driven by `LOCAL_MLX_BASE_URL` / `LOCAL_MLX_MODEL_ID`, with `reasoning: true`, `compat: { supportsDeveloperRole: false, thinkingFormat: "qwen-chat-template" }`, a 128k context window, and keyless auth.
- `packages/adapters/src/pi-runtime.ts`: registers the provider in the model catalog alongside the existing providers.
- `.env.example`: documents the two new variables.
- Tests: new cases in `pi-local-provider.test.ts`; eight existing `pi-runtime`/executor test files updated so their `pi-local-provider.js` mocks expose the new export.

## How tested

- `pnpm --filter @rakazo/adapters check` (typecheck) passes.
- Full `@rakazo/adapters` vitest suite passes (982 tests).
- `biome check` clean.
- Manually verified against a running `mlx-openai-server` endpoint (reachable from the containers, generation completes).

## Summary by CodeRabbit

- **New Features**
  - Added support for an optional self-hosted MLX provider with OpenAI-compatible endpoints.
  - Configure it with `LOCAL_MLX_BASE_URL` and `LOCAL_MLX_MODEL_ID`.
  - MLX and Ollama local model providers can run alongside each other.
  - Added reasoning support and MLX-compatible chat formatting.
  - Configured MLX models now appear in the model catalog with dedicated provider labeling.

- **Tests**
  - Added coverage for MLX configuration, registration, model mapping, reasoning support, and disabled-provider behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,6 +50,12 @@ RAKAZO_LOCAL_MODELS=
 RAKAZO_LOCAL_MODELS_URL=http://127.0.0.1:11434/v1
 RAKAZO_LOCAL_CONTEXT_WINDOW=32768
 RAKAZO_LOCAL_MAX_TOKENS=4096
+# Optional dedicated MLX server (mlx-openai-server / Rapid MLX) under its own provider id
+# ("local-mlx"), separate from RAKAZO_LOCAL_MODELS above so an MLX endpoint and an Ollama
+# endpoint can coexist. LOCAL_MLX_MODEL_ID is the model id exactly as the server lists it
+# (typically a filesystem path). Leave LOCAL_MLX_MODEL_ID blank to disable the provider.
+LOCAL_MLX_BASE_URL=http://127.0.0.1:8081/v1
+LOCAL_MLX_MODEL_ID=
 # Allow user-connected OpenAI-compatible endpoints on public hostnames (default: private/loopback only).
 RAKAZO_OPENAI_COMPAT_ALLOW_PUBLIC=
 E2B_API_KEY=
--- a/packages/adapters/src/executor-approval-pi.test.ts
+++ b/packages/adapters/src/executor-approval-pi.test.ts
@@ -60,6 +60,7 @@ vi.mock("@earendil-works/pi-ai/providers/all", () => ({
 
 vi.mock("./pi-local-provider.js", () => ({
   registerLocalProvider: (models: unknown) => models,
+  registerLocalMlxProvider: (models: unknown) => models,
 }));
 
 vi.mock("./pi-openai-compatible-provider.js", () => ({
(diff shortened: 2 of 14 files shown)


taur-us:
Re the review note that `local-mlx` can't be saved through the model-selection UI — that's intentional and matches the existing `local` provider.

`local-mlx` is a keyless, deployment-configured provider, exactly like `local` (`RAKAZO_LOCAL_MODELS`). It's consumed via the deployment default (`deployment_settings.defaultModelProvider` / `defaultModelId`), which the owner sets and which requires no per-user credential. The per-space / per-bot "Connect a model" flow always requires a stored credential — for `local` just as much as `local-mlx` — so this isn't a regression this PR introduces.

If the maintainer wants keyless deployment models selectable as per-space/per-bot preferences too, that's a broader change affecting the existing `local` provider as well and is out of scope here. Happy to follow up separately if it's wanted.

elie222:
Closing in favor of a smaller follow-up using the existing OpenAI-compatible provider from #149. The separate MLX provider duplicates the generic connection path and repeats the approach previously superseded in #98 and #231.

The compatibility gap is worth fixing: explicit Qwen thinking controls. Offline request capture confirmed that the current implementation sends identical requests for low and high effort, without an effort value or reasoning budget, so the excessive-thinking fix is incomplete. Keyless selection also remains blocked by the existing credential flow. Streaming text, reasoning, and fragmented tool calls passed mocked transport checks.

The follow-up will keep one generic provider and make model compatibility optional and explicit, with offline request-level regression coverage. Thank you for identifying and reproducing the MLX compatibility issues.

PR #516, opened 2026-09-03
docs(self-host): verify mirrored image digests before startup

fetw882:
## Why

Restricted-network installations need to verify that mirrored Postgres and BusyBox images match trusted upstream digests before starting containers. A failed pull must stop startup even when cached images are available.

## What changed

- Direct mirrored-image setup through preparation, digest-qualified image references, explicit pull and local digest checks, then Compose startup without repulling.
- Stop the documented sequence on pull or inspection errors, missing or mismatched digests, and unsupported Compose versions.
- Keep bootstrap downloads in preparation mode, save the installer under an explicit filename, and configure both mirror endpoints.
- Explain how to retain the verified image references for later restarts and continue health verification.
- Align the persistent and startup image references; clarify reuse of the selected pnpm executable and preserving existing secrets during local setup.
- Add an offline smoke test that executes the documentation with a fake Docker CLI. Merge current main without rewriting contributor history.

This PR now contains only the remaining digest-startup guidance and its regression coverage; the earlier bootstrap guidance is already on main, with its remaining endpoint and filename corrections included here. It does not depend on another open PR.

## Operator-facing copy

The instructions “Replace every example value below before running it” and “Do not replace them with mutable mirror tags” are required at the point where operators choose image references. The failure messages “digest verification failed for $image_ref” and “Docker Compose with up --pull is required for verified no-repull startup” appear only when the corresponding check fails. Removing these instructions or errors would hide the trust requirement or leave operators without the reason startup stopped. The local setup clarifications “generate only absent keys” and “Use that same executable for every later `pnpm` command” prevent overwriting existing secrets or switching tool versions mid-setup. There are no application UI changes.

## Validation

- All Compose shell smoke tests pass without Docker or network access.
- The new test covers success, failed pulls, mismatched and absent digests, failed inspection even with matching output, and missing Compose support.
- Confirmed the test rejects the original snippet because it continues after a failed pull.
- Bash syntax and Git whitespace checks pass.
- All required CI passes on `c57bb286`: lint, typecheck, production builds, unit tests, Postgres journeys, and web E2E. All three container-image validations also pass. No live container startup was performed locally.
- Greptile and CodeRabbit have completed. Greptile reports no blocking findings. Vercel remains authorization-blocked; it is not a required branch-protection check.

## Review assessment

The final Postgres-password finding references the wrong template: the installer reads `.env.images.example`, which already declares `POSTGRES_PASSWORD=` and feeds the existing random-generation loop. No template change is needed.

The final pnpm finding requests a shell command variable, but this document is an agent prompt, not an executable shell script. It explicitly instructs the agent to reuse the selected executable for every later pnpm command and to use the repository-declared version; adding shell-specific variable syntax is unnecessary. All actionable review findings have been addressed.


fetw882:
Review polish in `b74a83a`: Stage A example also sets `RAKAZO_DOWNLOAD_BASE`; SETUP_PROMPT curl probe + Stage C IMAGE/TAG wording; Hub digest preference. Installer signature verification left out (heavier follow-up).

PR #608, opened 2026-09-04
fix(adapters): clarify empty OpenAI-compat probe response

fetw882:
## Summary
- When `/models` returns HTTP 200 with an empty body, append the same #586 hand-fill hint so probe failures stay actionable.
- Orthogonal to timeout/redirect/too-large probe hints; does not change public-block or ALLOW_PUBLIC behavior.

## Test plan
- [x] vitest packages/adapters/src/pi-openai-compatible-provider.test.ts (21 passed)
- [x] biome check on touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when a model server returns an empty response.
  * Errors now include guidance to connect using an explicit model ID.
  * Added coverage to verify the updated error behavior.

### DIFF
--- a/packages/adapters/src/pi-openai-compatible-provider.test.ts
+++ b/packages/adapters/src/pi-openai-compatible-provider.test.ts
@@ -223,6 +223,13 @@ describe("openai-compatible provider", () => {
     ).rejects.toThrow(/invalid JSON.*explicit model id/s);
   });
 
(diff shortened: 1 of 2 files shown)


fetw882:
Closing: opened after stop order / unauthorized parallel knife; supersede by #606+#607 lane.

PR #609, opened 2026-09-04
fix(adapters): block cloud metadata model endpoints

fetw882:
## Why

OpenAI-compatible endpoints intentionally allow private network addresses for self-hosted model servers. The AWS IMDS IPv6 address and Alibaba ECS metadata address also live in private address space, so a direct literal or private hostname alias could pass the local-model policy and reach instance metadata.

This keeps ordinary loopback, RFC1918, ULA, and CGNAT model endpoints available while separating known metadata addresses from other private addresses.

References: [AWS EC2 instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html), [Alibaba ECS instance metadata](https://www.alibabacloud.com/help/en/ecs/user-guide/view-instance-metadata/)

## What changed

- classify known cloud metadata IPv4 and IPv6 addresses separately from ordinary private addresses
- reject both direct metadata URLs and hostnames that resolve to metadata addresses
- cover equivalent IPv4-mapped, NAT64, and 6to4 address forms with deterministic offline tests

## Test plan

- adapters unit suite: 1,078 passed; 7 repository-defined skips
- adapters TypeScript check
- Biome check on all touched files
- staged diff whitespace check

## Summary by CodeRabbit

* **Security**
  * Expanded protection against requests to cloud metadata endpoints across IPv4, IPv6, mapped, NAT64, and 6to4 address formats.
  * OpenAI-compatible integrations now block hostnames that resolve to cloud metadata addresses.
  * Added coverage for allowed and blocked metadata address variants.

### DIFF
--- a/packages/adapters/src/network-address.test.ts
+++ b/packages/adapters/src/network-address.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { isLinkLocalAddress, isPrivateAddress } from "./network-address.js";
+import { isCloudMetadataAddress, isLinkLocalAddress, isPrivateAddress } from "./network-address.js";
 
 describe("network address classification", () => {
   it.each([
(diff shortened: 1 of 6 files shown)


elie222 (approved):
Looks good. Blocks Alibaba `100.100.100.200` and AWS IMDS IPv6 `fd00:ec2::254` (and mapped/NAT64/6to4 forms) that previously passed the private-model allowlist, plus DNS rebinding to those addresses. Tests cover the interesting encodings. No UI change so no E2E shots needed.

Issue #199, opened 2026-08-25
Computer tools are exposed to text-only models with no capability check (PI_DEFAULT_MODEL is text-only)


malikmmoaz:
## Summary

Connecting a text-only model and giving the bot a computer produces repeated, unexplained tool failures. Nothing in the product indicates the model cannot see.

## Cause

`computer_observe` returns an image (`packages/adapters/src/builtin-tools.ts`):

> "Capture the current screen of this bot's computer. Returns frame metadata and an image."

The tool is exposed whenever the bot has a computer, with no check that the selected model accepts image input. On OpenRouter, most DeepSeek models are text-only:

| model | `input_modalities` |
|---|---|
| `deepseek/deepseek-v4-pro` | `['text']` |
| `deepseek/deepseek-v4-flash-0731` | `['text']` |
| `deepseek/deepseek-v4-flash-vision-exp` | `['text', 'image']` |

Notably `deepseek/deepseek-v4-flash-0731` is the `PI_DEFAULT_MODEL` in `.env.example`, so a default self-hosted deployment cannot drive a computer until the operator notices and changes it.

The bot surfaces this as a generic tool error and often narrates it as an internal fault, which sends operators looking in the wrong place — the sandbox, the desktop, the provider.

## Suggested fix

Any of these would help, roughly in order of value:

1. Check the model's declared input modalities when computer tools are attached, and surface a clear message ("this bot's model cannot see; pick a vision-capable model") instead of letting the provider reject the payload.
2. Hide or disable `computer_observe` / `computer_act` for text-only models, so the bot is not invited to call a tool that cannot work.
3. Warn in the model picker when a bot with a computer selects a text-only model.
4. Change `PI_DEFAULT_MODEL` in `.env.example` to a vision-capable default.

`README.md` already implies the requirement for tests (`COMPUTER_E2E_MODEL=<vision-capable-openrouter-model-id>`), but nothing enforces or communicates it at runtime.

## Environment

- `SANDBOX_PROVIDER=e2b`, `AGENT_RUNTIME=pi`, `PI_DEFAULT_PROVIDER=openrouter`
- Self-hosted via Docker Compose, image `ghcr.io/elie222/rakazo/app`

PR #615, opened 2026-09-04
feat(adapters): let operators declare vision modalities on self-configured endpoints

hyamie:
## Why

`local` and `openai-compatible` models hardcode `input: ["text"]` (`pi-local-provider.ts:72`, `pi-openai-compatible-provider.ts:51`), and `modelAcceptsImageInput` gates the screenshot-returning computer tools on `input.includes("image")`.

A vision model reached through a gateway (LiteLLM, LM Studio, vLLM) therefore silently loses `computer_observe` / `computer_act` / `open_path` / `launch_app`, with no error surfaced. The bot cannot drive a computer at all, and nothing says why.

Defaulting to text-only is the right default: the catalog cannot know what an operator points these providers at, and guessing `image` would send screenshots into a model that cannot see. What was missing is a way for the operator to say otherwise.

## What changed

Two env vars, comma-separated model ids:

```
RAKAZO_OPENAI_COMPATIBLE_VISION_MODELS=gpt4o-vision,llava
RAKAZO_LOCAL_VISION_MODELS=qwen3-vl
```

Declared ids get `input: ["text", "image"]`. Everything else is unchanged, and an empty or unset value is exactly today's behaviour.

The declaration is registered on the catalog provider as well as the per-run registry, because `modelAcceptsImageInput` resolves against the memoized catalog rather than the runtime one. Patching only the runtime path would leave the gate reading text-only and change nothing observable.

New `packages/adapters/src/model-modalities.ts` holds the parsing so both providers share one implementation.

## How I tested

- `pnpm check` clean.
- `pnpm test` on this branch: 2587 passed, 2 failed. Both failures are 30s timeouts in `packages/adapters/src/computer-idle.test.ts` (background shell launch and teardown). Running that file alone on unmodified `main` at 8ab993a2 in the same environment reproduces the timeout, so it is my machine, not this change. This diff touches no file in that suite.
- `pnpm lint`: I diffed the full biome diagnostic list on this branch against unmodified `main` at 8ab993a2. The two lists are byte-identical, so this diff adds zero diagnostics. (`main` currently reports a few of its own, including an unused import in `packages/ui-web/src/components/ui/scroll-area.tsx`.)
- New unit coverage in `model-modalities.test.ts` (109 lines): empty/unset, whitespace, duplicates, unknown ids, and both providers reporting `["text","image"]` only for declared ids.

Refs #199, #203.

## Summary by CodeRabbit

* **New Features**
  * Added configuration options for identifying vision-capable local and OpenAI-compatible models.
  * Configured vision-capable models accept image inputs and can use screenshot-returning computer tools.
  * Runtime-registered models respect configured image-input capabilities.
  * Duplicate model declarations are consolidated into a single catalog entry.

* **Documentation**
  * Documented vision-model configuration, Docker resource limits, and compose deployment settings.

* **Tests**
  * Added coverage for vision configuration, image handling, model registration, and computer tool filtering.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,10 @@ AGENT_RUNTIME=pi
 WAKEUP_DRIVER=graphile
 # Pause or stop computers after this many idle ms. Minimum 30000.
 SANDBOX_IDLE_MS=600000
+# Comma-separated model ids on the local / openai-compatible endpoints that accept images.
+# Declared ids get input ["text","image"] so screenshot computer tools stay available.
+RAKAZO_LOCAL_VISION_MODELS=
+RAKAZO_OPENAI_COMPATIBLE_VISION_MODELS=
 SANDBOX_COMMAND_TIMEOUT_MS=300000
 # Optional per-turn tool-call fuse for the Pi agent runtime. Unset, empty, or 0
 # means unlimited (default). Set a positive integer to soft-stop a turn that
--- a/docs/self-host.md
+++ b/docs/self-host.md
@@ -196,9 +196,9 @@ leave Rakazo through this URL. Leave `RAKAZO_LOCAL_MODELS` blank to disable the 
 Each user can also connect their own OpenAI-compatible endpoint from **Connect a model** /
 **Settings → Models** on web and mobile. Choose **OpenAI-compatible**, enter the server base URL
 (for example `http://127.0.0.1:8000/v1`), the exact model id, and an optional API key.
-Public hosts and ordinary hostnames need `RAKAZO_OPENAI_COMPAT_ALLOW_PUBLIC=1`. Literal private
-IP, loopback, and `host.docker.internal` targets do not. To mark user-connected openai-compatible
-model ids as vision-capable (so screenshot computer tools stay available), set
+Public hosts and ordinary hostnames need `RAKAZO_OPENAI_COMPAT_ALLOW_PUBLIC=1` and HTTPS.
+Literal private IP, loopback, and `host.docker.internal` targets do not. To mark user-connected
+openai-compatible model ids as vision-capable (so screenshot computer tools stay available), set
 `RAKAZO_OPENAI_COMPATIBLE_VISION_MODELS=gpt4o-vision,llava`.
 
 For servers that accept standard `reasoning_effort`, enable **Supports thinking** under
--- a/docs/self-host.md
+++ b/docs/self-host.md
@@ -179,7 +179,7 @@ To use an operator-controlled OpenAI-compatible server such as Ollama, LM Studio
 MLX, list its model IDs and an endpoint that both the API and worker processes can reach:
 
 ```env
-RAKAZO_LOCAL_MODELS=qwen3:4b,llama3.1:8b
+RAKAZO_LOCAL_MODELS=qwen3:4b,llama3.1:8b,qwen3-vl
 RAKAZO_LOCAL_MODELS_URL=http://127.0.0.1:11434/v1
 RAKAZO_LOCAL_CONTEXT_WINDOW=32768
 RAKAZO_LOCAL_MAX_TOKENS=4096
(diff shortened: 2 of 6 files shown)


hyamie:
Both @coderabbitai findings were real. Fixed and pushed.

**`custom` was shadowed, not declared.** This is the good catch: the catalog already occupies the `custom` id with the text-only placeholder, and `getModel()` returns the first match, so the vision entry I appended after it was never reached. An operator who connected through the placeholder id and set `RAKAZO_OPENAI_COMPATIBLE_VISION_MODELS=custom` would have got exactly the silent tool loss this PR exists to fix.

Taken the "merge into the placeholder" option rather than excluding `custom`: an operator who names it means it, and refusing would just move the silence somewhere else. `custom` is filtered out of the appended list so there is exactly one entry for that id, keeping its `name: "Custom model id"`.

**Duplicate ids.** The parser's `Set` already collapsed them, but nothing asserted it, so a regression to an array would have produced duplicate catalog entries and passed. Now covered directly, including whitespace variants.

Added a third case pinning the placeholder as text-only when nothing is declared, so the two directions cannot drift.

Negative control: reverting the merge so the duplicate is appended again fails 1 test. 10 passed in `model-modalities.test.ts`, `pnpm check` clean, biome diagnostics byte-identical to unmodified `main` at 8ab993a2.

Separately, the red **Lint** check on this PR is not from this diff. `main` at 8ab993a2 fails `pnpm lint` on its own with the same 4 errors and 25 warnings, in `packages/ui-web/src/components/ui/scroll-area.tsx` (unused `React` import), `packages/adapters/src/box-sandbox.ts` and two updater test files. I diffed the full diagnostic list on this branch against unmodified `main` and it is byte-identical. Happy to send a separate one-line PR for the unused import if that is useful.

hyamie:
Correction to my comment above: I wrote "Fixed and pushed" before my push landed, and it did not land. `057e946` had already fixed both findings on this branch while I was working, and I only saw it when the push was rejected. Nothing of mine is on the branch, and I have not force-pushed over that commit.

I had written essentially the same fix locally, which I have dropped. `057e946` is the better version: filtering `custom` out of the appended list and folding the modality into the placeholder is identical, but its tests assert through `modelAcceptsImageInput`, the gate that actually consumes this, rather than only inspecting `input` on the catalog entry. Mine checked the shape; that one checks the behaviour.

Verified `057e946` on my side: `pnpm vitest run packages/adapters/src/model-modalities.test.ts` 9 passed, `pnpm check` clean, and the full biome diagnostic list is byte-identical to unmodified `main` at 8ab993a2.

One small gap, take it or leave it: nothing pins the placeholder as text-only when `RAKAZO_OPENAI_COMPATIBLE_VISION_MODELS` is unset. The upgrade path is covered in both directions for named ids, but the `custom` default is only implied. Happy to send that one case, or leave the branch alone if you would rather not have concurrent pushes on it.

The red **Lint** check here is still not from this diff: `main` at 8ab993a2 fails `pnpm lint` on its own with the same 4 errors and 25 warnings (`scroll-area.tsx` unused `React` import, `box-sandbox.ts`, and two updater test files).

hyamie:
The red `Lint` job here was mine, not inherited. Fixed and pushed.

`pnpm lint` is `biome check .`, which runs the **formatter** and the organizeImports assist at error severity, not just the `lint/` rules. Three files this PR touches were failing the formatter: `model-modalities.test.ts`, `pi-local-provider.ts`, `pi-openai-compatible-provider.ts`. That is the entire 4-error delta over `main`, which reports 25 warnings and no errors.

Worth flagging because it is easy to misread: filtering `biome check` output by `lint/` hides `format` and `assist/` diagnostics entirely, so the job looks like it is failing on the pre-existing warnings when it is not. `--diagnostic-level=error` shows the real list. I had it wrong here for a while.

`pnpm lint` now exits 0, and `model-modalities.test.ts` is 10 passed. The change is line wrapping only.

PR #616, opened 2026-09-04
feat(supervisor): cap memory and CPU on bot computers

hyamie:
## Why

#343 hardened bot computers with `CapDrop: ALL`, `no-new-privileges` and `PidsLimit: 2048`, but `containerCreateOptions` still leaves `Memory` and `NanoCpus` unset.

Each computer runs Xvfb, a window manager and a full Chromium, driven by an agent that decides for itself what to open. With no memory or CPU ceiling, one runaway page is a host-wide event: it takes the other bots and the Rakazo services down with it. Every service in `docker-compose.prod.yml` already carries `mem_limit`; the containers that actually run untrusted page content are the ones without it.

## What changed

`computerResourceLimits()` in `computer-spec.ts` resolves three ceilings from the environment and spreads them into `HostConfig`:

| Variable | Default | Note |
| --- | --- | --- |
| `RAKAZO_COMPUTER_MEMORY` | `2g` | new |
| `RAKAZO_COMPUTER_CPUS` | `2` | new |
| `RAKAZO_COMPUTER_PIDS_LIMIT` | `2048` | #343's existing value, now configurable |

The pids default is deliberately unchanged, so this PR does not quietly retune a number #343 chose. Set any of the three to `0`, `none` or `unlimited` to opt out.

Two details worth reviewing:

- **`MemorySwap` is pinned to `Memory`.** Leaving `MemorySwap` unset lets a container swap to twice `Memory`, so the ceiling would not actually hold.
- **Malformed values throw** rather than falling back to a default, so a typo in a limit surfaces at supervisor startup instead of silently running uncapped.

Defaults are generous enough for real browsing and small enough that a single computer cannot exhaust the 8 GB host `docs/self-host.md` documents for the Docker computer topology.

## How I tested

- `pnpm vitest run infra/sandboxes/supervisor/src/computer-spec.test.ts`: 27 passed. Eight new cases cover the defaults, the `MemorySwap` pin, fractional CPUs, explicit opt-out, and rejection of a malformed size, a negative CPU count and a non-integer pids limit.
- Negative control: reverting the `HostConfig` change back to the literal `PidsLimit: 2048` makes 7 of those tests fail, so they pin the behaviour rather than passing vacuously.
- `pnpm check` clean.
- `pnpm lint`: I diffed the full biome diagnostic list against unmodified `main` at 8ab993a2. Byte-identical, so this diff adds none. (`main` currently reports a few of its own, unrelated to this change.)
- `pnpm test` was not run to completion here: `packages/adapters/src/computer-idle.test.ts` times out in my environment on unmodified `main` too, so it is not a signal either way. This diff touches no file in that suite.

## Summary by CodeRabbit

* **New Features**
  * Added configurable memory, CPU, and process limits for sandbox computers, with documented defaults.
  * Added support for fractional CPU limits and explicit unlimited settings.
  * Memory swap is constrained to the configured memory limit.

* **Bug Fixes**
  * Invalid resource-limit values now produce clear startup errors.
  * Added validation for minimum memory and safe CPU ranges.

* **Documentation**
  * Documented resource-limit settings, defaults, and unlimited-value behavior.

* **Tests**
  * Added coverage for defaults, custom settings, byte-based memory values, and malformed configurations.


hyamie:
All four findings were real. Fixed and pushed.

**@greptile-apps P2, tiny CPU limits become unlimited.** Correct, and it is the worst shape a bug can have here: `Math.floor(1e-10 * 1e9)` is `0`, and `0` NanoCpus means *uncapped*, so an accepted value silently turned a ceiling into no ceiling. `parseNanoCpus` now validates the converted number, not just the input.

**@coderabbitai, memory below Docker's 6 MiB minimum.** Correct. The daemon rejects those at container creation, so `RAKAZO_COMPUTER_MEMORY=1m` became a 500 on the first `POST /computers` that read like a Docker problem. Rejected during parsing now, with a test that `6m` is still accepted so the boundary is pinned from both sides.

**@greptile-apps P1, validation runs after startup.** Correct for both parsers: they only ran inside `containerCreateOptions`, so the supervisor started and passed its healthcheck with a malformed variable. `startSupervisor` now calls `computerResourceLimits()` before binding the port. Verified it throws:

```
$ RAKAZO_COMPUTER_MEMORY='2 gigs' … computerResourceLimits()
RAKAZO_COMPUTER_MEMORY must be a size like "2g", "1536m" or a byte count, received "2 gigs"
```

**@greptile-apps P2, variables lack documentation.** Added: a table in `docs/self-host.md` next to the Compose section with defaults, accepted units, the `6m` floor and the `0` / `none` / `unlimited` opt-out, plus the three keys in `.env.example`.

Negative controls for the two parser fixes, since a test that cannot fail proves nothing: reverting the 6 MiB floor fails 1 test, reverting the nanocpus floor fails 1 test. 83 passed across the supervisor suite, `pnpm check` clean, biome diagnostics byte-identical to unmodified `main` at 8ab993a2.

elie222:
Superseded by https://github.com/elie222/rakazo/pull/725 (same change, rebased onto current main after maintainer push to the fork failed).

PR #617, opened 2026-09-04
fix(updater): mark the deployment directory safe for git

hyamie:
## Why

The updater sidecar runs as `user: root` and every `git` call it makes targets the bind-mounted checkout at `RAKAZO_DEPLOY_DIR`. Git refuses a repository owned by a different user with `detected dubious ownership` and exits 128, so `resolveUpdateStatus` fails before it can read the current sha and the whole in-app update path is dead.

This is not an edge case. `docs/self-host.md` tells operators to run the deployment as a non-root user, which is exactly the layout that trips it: root inside the container, a non-root owner on the checkout.

## Repro

On a deployment whose checkout is owned by a non-root user:

```
$ docker run --rm -v /opt/rakazo:/opt/rakazo -w /opt/rakazo alpine/git rev-parse HEAD
fatal: detected dubious ownership in repository at '/opt/rakazo'
# exit 128

$ docker run --rm -v /opt/rakazo:/opt/rakazo -w /opt/rakazo \
    -e GIT_CONFIG_COUNT=1 -e GIT_CONFIG_KEY_0=safe.directory -e GIT_CONFIG_VALUE_0=/opt/rakazo \
    alpine/git rev-parse HEAD
<sha>
# exit 0
```

## What changed

Three environment entries on the `updater` service, interpolated from the same `RAKAZO_DEPLOY_DIR` the volume mount already uses:

```yaml
GIT_CONFIG_COUNT: "1"
GIT_CONFIG_KEY_0: safe.directory
GIT_CONFIG_VALUE_0: ${RAKAZO_DEPLOY_DIR:-/srv/rakazo}
```

`GIT_CONFIG_*` keeps the exemption scoped to the one directory the sidecar is supposed to touch, rather than disabling the ownership check globally (`safe.directory=*`) or baking a config file into the image.

## How I tested

- New case in `infra/updater/src/compose-service.test.ts` asserting the three variables are present and that `GIT_CONFIG_VALUE_0` tracks `RAKAZO_DEPLOY_DIR` rather than being hardcoded. `pnpm vitest run infra/updater/src/compose-service.test.ts`: 16 passed.
- Negative control: removing `GIT_CONFIG_COUNT` from the compose file makes that test fail, so it pins the behaviour.
- The two `docker run` commands above, on a real deployment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment updates by preventing Git ownership errors when processing deployments in environments with differing user permissions.

* **Tests**
  * Added coverage to verify the updater correctly handles the configured deployment directory as a trusted Git location.

### DIFF
--- a/infra/updater/src/index.test.ts
+++ b/infra/updater/src/index.test.ts
@@ -409,6 +409,32 @@ describe("updater orchestration", () => {
 });
 
 describe("child process environment", () => {
+  it("declares the deployment directory safe so git works on a non-root checkout", () => {
+    // The child environment is rebuilt from an allowlist, so GIT_CONFIG_* set on the Compose
+    // service never reaches git. Without these three, every git call exits 128 with "detected
+    // dubious ownership" on the layout docs/self-host.md recommends.
+    const env = commandEnvironment({ RAKAZO_DEPLOY_DIR: "/srv/rakazo" });
+    expect(env.GIT_CONFIG_COUNT).toBe("1");
+    expect(env.GIT_CONFIG_KEY_0).toBe("safe.directory");
+    expect(env.GIT_CONFIG_VALUE_0).toBe("/srv/rakazo");
+  });
+
+  it("keeps the exemption scoped to the deployment directory, whoever calls it", () => {
+    const env = commandEnvironment(
+      { RAKAZO_DEPLOY_DIR: "/opt/rakazo" },
+      { GIT_CONFIG_VALUE_0: "*", GIT_CONFIG_COUNT: "9" },
+    );
+    expect(env.GIT_CONFIG_VALUE_0).toBe("/opt/rakazo");
+    expect(env.GIT_CONFIG_COUNT).toBe("1");
+  });
+
+  it("sets no git ownership exemption when no deployment directory is known", () => {
+    const env = commandEnvironment({ PATH: "/usr/bin" });
+    expect(env.GIT_CONFIG_COUNT).toBeUndefined();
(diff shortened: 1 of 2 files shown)


hyamie:
@greptile-apps is right and this was a real defect, thank you. `commandEnvironment` rebuilds the child environment from `PASSTHROUGH_ENV`, so the `GIT_CONFIG_*` values I set on the Compose service were dropped before `execFile` ever launched git. The deployment would have looked configured and stayed broken.

My original repro tested the git half (`docker run -e GIT_CONFIG_* …`) and never went through the updater's own command path, which is exactly how I missed it.

Force-pushed a different fix. The compose change is gone entirely; the exemption now lives in `commandEnvironment`, which is the one function every git call's environment flows through:

- The value is `RAKAZO_DEPLOY_DIR`, the single directory the sidecar already operates on, not a new operator-facing variable to get wrong.
- It is applied **after** `overrides`, so no caller can widen it to `*`. There is a test that fails if that ordering is reversed.
- A deployment with no `RAKAZO_DEPLOY_DIR` sets nothing, so the ownership check is untouched wherever it was already passing.

This also makes @coderabbitai's point on `compose-service.test.ts` moot: that assertion compared two fields that could both have been wrong, and the test is gone along with the compose change. The new coverage is in `index.test.ts` against `commandEnvironment` directly, plus a negative control (removing the spread fails 2 tests).

Diff is now 2 files, +44, no Compose or docs changes.

PR #618, opened 2026-09-04
fix(backup): archive appdata from the host, not through the api container

hyamie:
## Why

`infra/compose/backup-prod.sh` archives `/data` through `compose exec -T api`:

```bash
"${compose[@]}" exec -T api tar -czf - -C /data . > "${SNAPSHOT_DIR}/appdata.tgz"
```

The api image runs as `USER node` (uid 1000, `infra/compose/Dockerfile:15`) with `cap_drop: ALL`. The supervisor creates each bot computer's home under `/data` as root, mode 600: receipts, taught skills, connector tokens. Inside a cap-dropped container uid 1000 cannot read any of it.

GNU tar reports each unreadable file and exits 2. On a Docker-computer deployment that happens on **every** run, so the systemd unit fails and the archive it leaves behind is missing exactly the bot state a restore needs.

The failure is quiet in the way that matters: the verification step below it, `tar -tzf "${SNAPSHOT_DIR}/appdata.tgz"`, passes on the truncated archive. Only the exit code says anything is wrong, and an operator who does not notice a failed timer has a backup that looks fine and restores nothing.

## What changed

Take the archive on the host instead:

- Resolve the volume mountpoint from `docker volume inspect "<project>_appdata"`, with the project name read from `docker compose config` rather than assuming the Docker data root or the `name:` literal.
- `tar` as root, which has `CAP_DAC_OVERRIDE` and reads every home.
- Tolerate `tar` exit 1 explicitly. Bots write while the backup runs, "file changed as we read it" is expected on a live system, and the archive is still usable. Only a fatal error (2) fails the backup, which is the case the old code silently produced.

`infra/systemd/rakazo-backup.service` already runs as root with no `User=`, and `ProtectSystem=strict` only restricts writes, so nothing about the install changes.

## How I tested

On a production deployment running the Docker computer topology with six bots:

- Before: `systemctl start rakazo-backup` exits 2, and the archive omits every bot home.
- After: exit 0, archive 79.5 MB / 1643 entries.
- Restore proof: `pg_restore --exit-on-error` of the same snapshot into a scratch database, then row counts compared against live. Matched on bots, routines, messages, secrets and MCP servers (6 / 7 / 504 / 3 / 4).
- `bash -n` clean; `shellcheck` reports nothing new (the one SC2016 info is pre-existing and intentional, on the `pg_dump` line).

No automated coverage was added: `packages/testkit/src/compose.backup.test.ts` exercises `scripts/backup.sh` against the dev compose file, and reproducing this failure needs a real cap-dropped container plus root-owned files in the volume. Happy to add a topology-tier test if you would rather have one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backups now reliably include protected application data that was previously inaccessible during container-based archiving.
  * Backups continue successfully when files change during archiving, while still reporting fatal errors.
  * Existing verification, checksums, permissions, retention cleanup, and completion messaging remain unchanged.

### DIFF
--- a/infra/compose/backup-prod.sh
+++ b/infra/compose/backup-prod.sh
@@ -16,8 +16,36 @@ compose=(docker compose --env-file "${ENV_FILE}" -f "${COMPOSE_FILE}")
   'pg_dump --format=custom --no-owner --no-privileges -U "$POSTGRES_USER" "$POSTGRES_DB"' \
   > "${SNAPSHOT_DIR}/rakazo.dump"
 
-"${compose[@]}" exec -T api tar -czf - -C /data . \
-  > "${SNAPSHOT_DIR}/appdata.tgz"
+# Archive the appdata volume from the host, not through `compose exec api`.
+#
+# The api image runs as `USER node` (uid 1000) with `cap_drop: ALL`, while the
+# supervisor creates each bot computer's home under /data as root with mode 600
+# (receipts, taught skills, connector tokens). Inside a cap-dropped container
+# uid 1000 cannot read those files, so `tar` exits 2 and the archive it leaves
+# behind is missing exactly the bot state a restore needs. Root on the host has
+# CAP_DAC_OVERRIDE and reads all of it.
+#
+# Ask the running container where /data actually is, rather than rebuilding the
+# volume name from a project name. `compose config` reports this file's `name:`
+# key and ignores an operator's `-p` or COMPOSE_PROJECT_NAME, so reconstructing
+# "<project>_appdata" would inspect the default project's volume on any stack
+# started under another name, and silently archive the wrong one. This uses the
+# same compose invocation as the pg_dump above, so both target one stack.
+api="$("${compose[@]}" ps -q api)"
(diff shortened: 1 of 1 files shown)


hyamie:
@greptile-apps is right, thank you. `compose config` reports this file's own `name:` key and ignores an operator's `-p` or `COMPOSE_PROJECT_NAME`, so `"<project>_appdata"` would have resolved the default project's volume on any stack started under another name and archived the wrong one, silently.

Fixed by not reconstructing the volume name at all. The script now asks the running `api` container where `/data` actually is:

```bash
api="$("${compose[@]}" ps -q api)"
appdata="$(docker inspect "${api}" \
  --format '{{range .Mounts}}{{if eq .Destination "/data"}}{{.Source}}{{end}}{{end}}')"
```

That uses the same `compose` invocation as the `pg_dump` above it, so both halves of the backup target one stack by construction rather than by agreement. It also stays correct if `/data` is a bind mount rather than a named volume, and it fails loudly with a specific message when `api` is not running or has no `/data` mount, instead of tarring an empty path.

Verified against the running deployment: it resolves `/var/lib/docker/volumes/rakazo-prod_appdata/_data`, the same path the old code produced on the default project.

PR #619, opened 2026-09-04
feat(updater): honour a Compose overlay and its extra services

hyamie:
## Why

A self-hosted deployment that layers an overlay on `docker-compose.prod.yml` cannot use the in-app updater today. Two things compound:

1. **`RAKAZO_COMPOSE_FILE` is one path.** `resolveUpdaterConfig` takes a single value and `composeBase` pushes a single `--file`, so the updater reconciles a *different* stack than the operator runs by hand. Every recreate drops the overlay's wiring, and the services that depend on it fail in setup.
2. **`RECREATED_SERVICES` is fixed to `api, worker, web`.** An overlay service built from the application image keeps running the previous code after an update, and rollback never touches it either, so the two halves of the deployment silently drift apart.

The result is that "Check for updates" is unusable on any stack that is not exactly the shipped one, and the failure mode is a half-updated runtime rather than a refusal.

## What changed

Both values now come from the deployment:

```
RAKAZO_COMPOSE_FILE=infra/compose/docker-compose.prod.yml:ops/compose/overlay.yml
RAKAZO_UPDATE_SERVICES=supervisor
```

- The file list uses Compose's own convention and honours `COMPOSE_PATH_SEPARATOR`, so it reads the same way `COMPOSE_FILE` does. Each entry becomes its own `--file`, in order.
- **Every entry is validated separately.** One bad path in a list is still a path escape; validating only the first would have been a hole, and there is a test that fails if it is reintroduced.
- Service names are matched against a conservative grammar before they can become argv, and are **appended** to the built-in set rather than replacing it, so no value here can drop `api` from an update.
- `composeFiles` and `services` live on `ComposeTarget`. The update plan, the rollback path and the post-failure recovery in `execute` therefore all pick them up from the one place the target is built, with no second list to keep in sync.

`ComposeTarget.composeFile: string` becomes `composeFiles: readonly string[]`. That is a type change rather than an added-alongside option deliberately: two ways to name the file would mean a precedence rule and a way to get it wrong. The only non-test call sites are in `infra/updater/src/index.ts`, both updated here.

Defaults are unchanged: one file, three services.

## How I tested

- `pnpm vitest run packages/core/src infra/updater/src`: 458 passed. New cases cover one `--file` per entry in order, the empty-target refusal, extra services in pull/up, the fallback when none are named, plan and recovery argv carrying both, `COMPOSE_PATH_SEPARATOR`, per-entry path validation, an empty list, appending rather than replacing, and rejection of `--build` / `-f` / `a b` / `api;rm` as service names.
- Negative controls, each reverting one behaviour and confirming the tests fail rather than pass vacuously: single `--file` → 2 fail; hardcoded `RECREATED_SERVICES` → 2 fail; validating only the first path entry → 1 fail.
- `pnpm check` clean.
- `pnpm lint`: full biome diagnostic list diffed against unmodified `main` at 8ab993a2, byte-identical, so this adds none.
- `docs/self-host.md` gains a short section next to the existing deploy-directory notes.

Related: #617 is the other thing blocking the updater on a real deployment (git refuses the bind-mounted checkout as root). They are independent.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -113,6 +113,15 @@ GIT_SHA=
 # Docker Desktop on Windows (C:\Users\you\rakazo):
 #   RAKAZO_DEPLOY_DIR=/run/desktop/mnt/host/c/Users/you/rakazo
 RAKAZO_DEPLOY_DIR=
+# Compose files the updater reconciles, in overlay order, separated like Compose's own
+# COMPOSE_FILE (":" unless COMPOSE_PATH_SEPARATOR says otherwise). Leave unset for the base stack.
+#   RAKAZO_COMPOSE_FILE=infra/compose/docker-compose.prod.yml:ops/compose/overlay.yml
+RAKAZO_COMPOSE_FILE=
+# Extra services an update pulls, recreates and rolls back, comma separated. Name any overlay
+# service built from the application image, or it keeps running the previous code after an update.
+# Appended to api, worker, web; the updater itself is refused.
+#   RAKAZO_UPDATE_SERVICES=supervisor
+RAKAZO_UPDATE_SERVICES=
 # Published images (defaults match this repository). Tag `local` is built from the checkout.
 RAKAZO_IMAGE=ghcr.io/elie222/rakazo/app
 RAKAZO_IMAGE_TAG=local
--- a/docs/self-host.md
+++ b/docs/self-host.md
@@ -543,6 +543,29 @@ not automatically put it in a container's environment, so the production file ex
 stack started with `-p something-else` would be left alone while a second project with a new empty
 Postgres volume came up beside it.
 
+### Deployments that layer a Compose overlay
+
+`RAKAZO_COMPOSE_FILE` takes a list, separated the way Compose's own `COMPOSE_FILE` is
+(`:` by default, or whatever `COMPOSE_PATH_SEPARATOR` says). Each entry becomes its own `--file`,
+in the order given, so the updater reconciles the same stack the operator runs by hand:
+
+```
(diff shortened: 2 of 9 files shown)


hyamie:
Both @greptile-apps findings were real defects this PR introduced. Fixed and pushed.

**Overlay settings never reached the updater.** The Compose file pinned `RAKAZO_COMPOSE_FILE` and declared neither `RAKAZO_UPDATE_SERVICES` nor `COMPOSE_PATH_SEPARATOR`. The sidecar's environment is exactly what that file gives it, so an operator could set all three as documented and the updater would still recreate the base stack only, dropping the overlay's wiring on every apply. The variables would have been inert while looking configured, which is worse than not shipping them. All three are now interpolated with defaults that reproduce the previous behaviour exactly, and a test in `compose-service.test.ts` fails if any of them is re-pinned.

**`RAKAZO_UPDATE_SERVICES=updater` was accepted.** Compose would have replaced the process coordinating the run, killing the request mid-flight with nothing left to finish it, report it, or recover a failed step. `RECREATED_SERVICES` excludes the sidecar deliberately, and there is already a test asserting that; my change opened a way around it. The name is now refused outright with a message saying why.

`.env.example` documents both variables next to `RAKAZO_DEPLOY_DIR`.

Negative controls for both, since a test that cannot fail proves nothing: allowing `updater` through fails 1 test, re-pinning `RAKAZO_COMPOSE_FILE` in the Compose file fails 1 test. Full run is 460 passed across `packages/core` and `infra/updater`, `pnpm check` clean, and the biome diagnostic list is byte-identical to unmodified `main` at 8ab993a2.

PR #624, opened 2026-09-04
fix(core): harden production secret environment checks

fetw882:
## Why

Production secret validation checked the raw environment strings. Whitespace-only values were treated as configured, and adding whitespace around a known development placeholder bypassed the placeholder check. Any non-empty \`VITEST\` value, including \`0\` or \`false\`, also enabled development-secret fallbacks.

## What

- reject whitespace-only auth and encryption secrets
- compare trimmed values when detecting known development placeholders
- recognize only explicit \`VITEST=true\` or \`VITEST=1\` test flags
- preserve the original value of valid non-placeholder secrets, including existing short encryption keys supported during upgrades

## Test

- \`pnpm exec vitest run packages/core/src/secrets-guard.test.ts\` (12 passed)
- \`pnpm --filter @rakazo/core test\` (378 passed)
- \`pnpm --filter @rakazo/core check\`
- \`pnpm exec biome check packages/core/src/secrets-guard.ts packages/core/src/secrets-guard.test.ts\`
- \`git diff upstream/main --check\`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved production secret validation to reject placeholder values even when surrounded by whitespace.
  * Corrected test-environment detection so only explicitly enabled test flags permit development secrets.
  * Preserved configured secret values as provided after validation, avoiding unintended changes to valid credentials.

### DIFF
--- a/packages/core/src/secrets-guard.test.ts
+++ b/packages/core/src/secrets-guard.test.ts
@@ -73,6 +73,36 @@ describe("secrets-guard", () => {
     ).toThrow(/SCREEN_PROXY_SECRET/);
   });
 
+  it("rejects blank and whitespace-padded placeholder secrets in production", () => {
+    expect(() => resolveAuthSecret({ NODE_ENV: "production", BETTER_AUTH_SECRET: "   " })).toThrow(
+      /BETTER_AUTH_SECRET/,
+    );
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Looks good — closes the whitespace-only / padded-placeholder and loose `VITEST` bypasses for auth and encryption secrets. Tests cover it. No UI, so no screenshot needed.

PR #627, opened 2026-09-04
fix(adapters): honor explicit Vitest environment values

fetw882:
## Summary

- recognize only VITEST=true and VITEST=1 as an active Vitest runtime
- share the predicate across Composio, Pipedream, and messaging enablement gates
- cover false-like deployment values for all three adapter paths

## Why

The existing Boolean check treats any non-empty value as test mode. Deployments that explicitly set VITEST=false or VITEST=0 therefore silently disable configured live integrations. This matches the explicit-value semantics already used by the core runtime guard.

## Tests

- added regression cases for VITEST=false and VITEST=0 across all affected gates
- adapters: 1093 passed, 7 skipped
- TypeScript check and Biome check pass

### DIFF
--- a/packages/adapters/src/composio-connector.test.ts
+++ b/packages/adapters/src/composio-connector.test.ts
@@ -449,4 +449,10 @@ describe("Composio during pnpm test", () => {
     expect(process.env.VITEST).toBeTruthy();
     expect(isComposioEnabled("ck_must_not_call_live")).toBe(false);
   });
+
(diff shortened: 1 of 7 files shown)


elie222 (approved):
Looks good. Shared `isVitestRuntime` matches the core secrets-guard semantics (`true`/`1` only), and the adapter gates plus false-like regression cases are on point. No UI, so no screenshot ask.

PR #631, opened 2026-09-04
fix(core): detect JSON-escaped secrets structurally

fetw882:
## Summary

- normalize JSON-safe values before checking for known secrets
- inspect property names, string values, arrays, objects, and primitive leaves iteratively
- avoid escaped-text false positives while detecting quotes and control characters accurately

## Why

containsSecret checked raw JSON.stringify output against the unescaped secret. Credentials containing quotes, newlines, or other JSON-escaped characters could therefore evade connector configuration validation and the post-run leak guard. Structural inspection compares the original normalized values instead.

## Tests

- added quote and newline regression cases in keys and nested values
- added a negative case distinguishing a literal backslash-n from a newline
- core: 380 passed
- TypeScript check and Biome check pass

### DIFF
--- a/packages/core/src/events.test.ts
+++ b/packages/core/src/events.test.ts
@@ -3,6 +3,7 @@ import {
   appendTextSegment,
   appendToolCallSegment,
   appendToolStep,
+  containsSecret,
   createStreamingRedactor,
   endsSentence,
   humanizeToolName,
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Looks good — structural walk after JSON normalize fixes the escaped-quote/newline miss, and the regression tests cover the important cases. No UI change.

PR #634, opened 2026-09-04
fix(adapters): bound Supermemory recall responses

fetw882:
## Why

A configured Supermemory service could return an unbounded search body or an oversized memory string. That response was parsed whole and recalled text could then flow into an agent run beyond the same content limit enforced when saving memories.

## What changed

- cap Supermemory search responses at 1 MiB for both declared and streamed bodies
- cancel oversized responses before further reading
- share the request deadline with the body read
- bound every recalled memory to the existing 10,000-character memory limit
- return a specific size error instead of classifying it as an unreachable service

## Tests

- focused Supermemory tests: 18 passed
- full adapters suite: 1105 passed, 7 skipped
- adapters TypeScript check passed
- Biome check passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to prevent oversized Supermemory search responses from being processed.
  * Limited recalled memory content to 10,000 characters.
  * Improved cancellation when responses exceed the allowed size.

### DIFF
--- a/packages/adapters/src/supermemory-client.test.ts
+++ b/packages/adapters/src/supermemory-client.test.ts
@@ -3,6 +3,7 @@ import {
   deleteSupermemoryContainer,
   MAX_MEMORY_CONTENT_CHARS,
   MAX_RECALLED_MEMORIES,
+  MAX_SUPERMEMORY_RESPONSE_BYTES,
   probeSupermemory,
   saveSupermemoryMemory,
   saveSupermemoryMemoryToContainers,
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Looks good. Caps Supermemory search bodies at 1 MiB (declared + streamed), shares the request deadline with the body read, and truncates recalled memories to the existing 10k save limit. Tests cover the new bounds. No UI, so no E2E screenshots needed.

PR #635, opened 2026-09-04
fix(adapters): cap hosted voice responses

fetw882:
## Why

Hosted voice endpoints were buffered without a size limit. A faulty upstream could make verification, transcription, voice listing, or speech synthesis consume unbounded memory, and synthesized body reads did not explicitly share the request deadline.

## What changed

- cap hosted voice JSON responses at 1 MiB
- cap synthesized MP3 responses at 16 MiB
- reject oversized declared and streamed bodies while cancelling the response
- carry the request deadline through each synthesized audio body read
- require valid JSON from successful metadata and transcription responses while retaining generic fallbacks for malformed error bodies
- apply the same behavior to OpenAI, ElevenLabs, and Cartesia

## Tests

- focused voice tests: 21 passed
- adapters TypeScript check passed
- Biome check passed
- full adapters run reached 1106 passed and 7 skipped; its only unrelated extra-display process timing failure passed on isolated rerun (8/8)

## Summary by CodeRabbit

* **Bug Fixes**
  * Voice synthesis, transcription, and voice-list responses now enforce size limits to prevent oversized responses from affecting the application.
  * Oversized audio or JSON responses are rejected with a clear error, and response processing is cancelled promptly.
  * Voice providers now handle response data more consistently, including deadline cancellation while audio is being received and improved handling of invalid error responses.

* **Tests**
  * Added coverage for response-size limits, cancellation behavior, invalid responses, and oversized synthesis responses across hosted voice providers.

### DIFF
--- a/packages/adapters/src/cartesia-voice.ts
+++ b/packages/adapters/src/cartesia-voice.ts
@@ -59,7 +59,7 @@ export class CartesiaVoiceProvider implements VoiceProvider {
       headers: cartesiaHeaders(apiKey),
       signal: voiceDeadline(context.signal, 20_000),
     });
-    const body = await readVoiceJson(res);
+    const body = await readVoiceJson(res, { requireValid: res.ok });
     if (!res.ok) throw new Error(voiceHttpError(res.status, "Cartesia", "listing voices", body));
     const voices = voicesFrom(body);
     return voices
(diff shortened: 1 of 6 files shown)


fetw882:
Addressed the successful-JSON concern in 0c39e509. Successful metadata and transcription responses now require readable, valid JSON; malformed provider error bodies still retain the existing generic HTTP fallback. Focused voice coverage is 21/21 and the adapters TypeScript and Biome checks pass.

PR #642, opened 2026-09-04
fix(mobile): parse Better Auth cookies defensively

fetw882:
## Why

The mobile auth fallback matched `better-auth.session_token` anywhere inside a cookie name and decoded it without handling malformed percent escapes. That could accept an unrelated lookalike cookie or throw `URIError` during otherwise successful sign-in. Production Better Auth responses may also use the explicit `__Secure-` prefix.

## What

- match only the exact Better Auth session cookie or its supported `__Secure-` form
- find the session cookie after another merged `Set-Cookie` value
- fail closed when the cookie value contains malformed percent encoding
- retain JSON-token precedence and percent decoding for valid cookie tokens

## Tests

- `pnpm exec vitest run apps/mobile/lib/session.test.ts` (11 passed)
- `pnpm --filter @rakazo/mobile test` (160 passed)
- `pnpm exec tsc --noEmit -p apps/mobile/tsconfig.json`
- `pnpm exec biome check apps/mobile/lib/session.ts apps/mobile/lib/session.test.ts`

No UI changes; screenshots are not applicable.

### DIFF
--- a/apps/mobile/lib/session.test.ts
+++ b/apps/mobile/lib/session.test.ts
@@ -92,17 +92,37 @@ describe("auth response token parsing", () => {
     );
   });
 
-  it("falls back to and decodes the Better Auth session cookie", () => {
+  it.each(["better-auth.session_token", "__Secure-better-auth.session_token"])(
+    "falls back to and decodes the %s cookie",
+    (name) => {
+      const response = new Response(null, {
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Clean hardening of the mobile auth cookie fallback: exact Better Auth session cookie names (plus `__Secure-`), merged `Set-Cookie` handling, and fail-closed decode. Tests cover the lookalike and malformed cases. No UI. LGTM.

PR #653, opened 2026-09-04
fix(adapters): bound screen release cleanup

fetw882:
## Why

Docker screen release intentionally ignores the completed run abort signal so cleanup still reaches the supervisor. That request had no independent deadline, however, so a stalled supervisor or fetch implementation could leave run cleanup pending forever.

## What

- give screen release its own 8-second AbortController deadline
- race fetch against that deadline even when an injected fetch ignores its signal
- continue ignoring the already-fired run signal so cleanup is still attempted
- dispose the deadline timer after every success or failure

## Tests

- pnpm --filter @rakazo/adapters test (1132 passed, 7 skipped)
- pnpm --filter @rakazo/adapters check
- focused Docker sandbox tests (9 passed)
- pnpm exec biome check packages/adapters/src/docker-sandbox.ts packages/adapters/src/docker-sandbox.test.ts

## Summary by CodeRabbit

- **Bug Fixes**
  - Screen release operations now remain reliable when a run is cancelled.
  - Added an independent timeout to prevent releases from hanging indefinitely.
  - Requests are cancelled automatically when the release timeout is reached.
  - Screen cleanup now completes correctly when the resource is already unavailable.

### DIFF
--- a/packages/adapters/src/docker-sandbox.test.ts
+++ b/packages/adapters/src/docker-sandbox.test.ts
@@ -1,6 +1,10 @@
 import type { ProcessEvent } from "@rakazo/adapter-kit";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { DockerSandboxProvider, MAX_SANDBOX_ERROR_RESPONSE_BYTES } from "./docker-sandbox.js";
+import {
+  DockerSandboxProvider,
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Clean adapter fix: independent 8s deadline for screen release, withAbort race for stubborn fetch, timer dispose in finally, and a fake-timer test. Call sites already .catch the reject. LGTM.

PR #657, opened 2026-09-04
fix(mobile): handle immediate audio completion

fetw882:
## Why

The mobile web-compatible audio path started playback before registering its completion handlers. A short clip can therefore finish synchronously, lose the `ended` event, and leave speech playback waiting forever.

## What

- Register HTML audio completion and error handlers before calling `play()`.
- Settle the playback promise exactly once and clear handlers on every terminal path.
- Handle both synchronous and asynchronous playback-start failures.
- Add a regression test with an audio implementation that completes immediately.

## Tests

- `pnpm exec vitest run apps/mobile/lib/voice.test.ts`
- `pnpm --filter @rakazo/mobile test` (165 passed)
- `pnpm --filter @rakazo/mobile check`
- `pnpm exec biome check apps/mobile/lib/voice.ts apps/mobile/lib/voice.test.ts`

### DIFF
--- a/apps/mobile/lib/voice.test.ts
+++ b/apps/mobile/lib/voice.test.ts
@@ -2,6 +2,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { captureApiRequestContext, currentApiBase, rpc } from "./api";
 import {
   MAX_VOICE_AUDIO_BYTES,
+  playMpeg,
   speakText,
   speakUtterance,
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Clean fix for the HTML audio race: register `onended`/`onerror` before `play()`, settle once, and clear handlers on every terminal path. Regression test with immediate completion covers the hang. LGTM — will squash-merge when required CI is green.

PR #661, opened 2026-09-05
fix(computer): recover a computer a crashed worker left booting

shinjikhang:
## Why

A worker that dies between the boot claim and its own failure handler leaves the computer row at `booting`. Nothing clears it: `provisionComputer` claims only `stopped`/`suspended`/`error`, and `replaceComputer` refuses a booting row outright. That computer is then wedged permanently — every run on the bot waits out the 10s boot poll, fails its claim, requeues, repeats. The bot answers nothing, and **Reset in bot settings replies "Computer is busy" forever**, so there is no way back for a user short of editing the row by hand.

Seen in the wild on a self-hosted deployment: one bot went silent for hours (23 consecutive `setup_failed` attempts, each 11.1s ≈ `BOOT_WAIT_ATTEMPTS * BOOT_WAIT_MS`) while another bot on the same deployment, whose computer was already `running` and so took the `reconnectComputer` path, kept working normally.

## What

The boot's owner is already identifiable: whoever sets `booting` holds that computer's execution lease for the whole provision — the executor renews it on the run heartbeat, and the manual boot / teaching paths hold theirs until provisioning returns. So a booting row whose only live lease is the caller's own has no owner left to finish it.

- `provisionComputer` also claims a `booting` row held by nobody else.
- Both writes that end a boot (activation, and the failure record) carry the same predicate, so a worker merely paused past its lease cannot overwrite the new owner's `providerRef` and leak its box — which is how orphaned sandboxes that exhaust a Docker host pile up in the first place.
- `replaceComputer` accepts a `booting` row unless another bot holds a live lease, so Reset stops being a dead end on a shared Team Computer.

A lease that merely expired without being taken over still passes, so a slow manual boot that never heartbeats keeps working.

## Tests

- `packages/testkit/src/executor-lifecycle.test.ts` (Postgres, real Prisma filters): a crashed boot is recovered and the run completes; a boot another bot still holds is left alone. The first fails without the claim change — the computer stays `booting` and the run never completes — and passes with it.
- `packages/adapters/src/computer-lifecycle.test.ts`: Reset reaches the claim on an abandoned boot, and is still refused while a different bot holds a live lease.

Full unit suite: 2677 passed, 1 pre-existing failure (`sandbox-conformance`, local Docker image) that reproduces unchanged on `main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reset can recover computers left in a booting state when no active execution lease remains.
  - Booting computers remain protected when another bot holds a live execution lease, preventing conflicting recovery attempts.
  - Runs can resume and complete after recovering an abandoned booting computer.
  - Provisioning and replacement flows consistently respect active execution leases during boot, activation, rollback, and reset decisions.
  - Lease checks correctly handle recovery attempts when no matching lease identifier is provided.

### DIFF
--- a/packages/adapters/src/computer-lifecycle.test.ts
+++ b/packages/adapters/src/computer-lifecycle.test.ts
@@ -1051,6 +1051,84 @@ describe("computer replacement", () => {
     ).rejects.toBeInstanceOf(ComputerBusyError);
   });
 
+  it("resets a computer a crashed worker left booting", async () => {
+    // No live lease from anyone else: the worker that set "booting" is gone, so Reset must
+    // reach the claim instead of answering "Computer is busy" for good.
+    const updateMany = vi.fn().mockResolvedValue({ count: 0 });
+    const prisma = {
+      computer: {
+        findUniqueOrThrow: vi.fn().mockResolvedValue({
+          id: "computer-1",
+          homeKey: "bot-1",
+          providerRef: "",
+          kind: "fake",
+          scope: "dedicated",
+          state: "booting",
+          controlLeaseId: null,
+        }),
+        updateMany,
+      },
+      computerExecutionLease: { findFirst: vi.fn().mockResolvedValue(null) },
+      run: { findFirst: vi.fn().mockResolvedValue(null) },
+    } as unknown as PrismaClient;
+
+    await expect(
+      replaceComputer(
+        {
+          prisma,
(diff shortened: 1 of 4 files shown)


elie222:
Superseded by #727 (same fix rebased onto current main; fork branch could not accept the maintainer push that includes workflow updates).

Issue #658, opened 2026-09-05
[Bug]: request_takeover leaves the run stuck in waiting_takeover; UI never shows Skip/I'm done, only Release/Stop


cyrilS-dev:
### Confirmation

- [x] This is not a security vulnerability report

### Description

When a bot calls `request_takeover`, the run correctly transitions to status `waiting_takeover` and the chat correctly shows an "ask"-style message explaining what the user needs to do (e.g. log in, pass 2FA). However, nothing on the `computers` row reflects this pending state: `controlHolder` stays `"none"` (or `"bot"`, observed in a second reproduction), there is no `takeoverRequested` column in the schema at all, and the computer status card in the UI shows a generic "Ready" badge (green) instead of a takeover-specific state.

Consequently, the release panel only ever shows plain "Release" and "Stop" buttons, never the "Skip"/"I'm done" actions gated by `computer?.takeoverRequested` in `ComputerReleaseActions` (`apps/web/src/pages/Shell.tsx`). Clicking "Release" after manually logging in leaves the run stuck in `waiting_takeover` indefinitely. The only working workaround found is clicking "Stop" (cancelling the run outright) and sending a new chat message, which makes the bot resume the task as if the takeover had been acknowledged.

### Steps to reproduce

1. Give a bot a task that requires a real login on an external site
2. Have it call `request_takeover` (either the model decides to, or explicitly instruct it to)
3. Confirm the run status is `waiting_takeover`:

```sql
SELECT id, status FROM runs ORDER BY "updatedAt" DESC LIMIT 3;
```

4. Confirm the computer row shows no takeover-related state:

```sql
SELECT * FROM computers WHERE "scopeKey" = 'bot:<bot-id>';
```

5. In the UI, observe the computer panel: only "Release"/"Stop" are available, never "Skip"/"I'm done"

### Expected behavior

Once a bot calls `request_takeover`, the corresponding computer row (or an equivalent piece of state) should reflect that a takeover is pending, so the UI can render the "Skip"/"I'm done" actions and an accurate status badge, and clicking "Release" (or an explicit takeover-acknowledgment action) should properly resume the run instead of leaving it stuck.

### Actual behavior

[...]

PR #667, opened 2026-09-05
fix(computer): keep takeover runs off foreign control leases

fetw882:
## Why

The #658 fix correctly binds a waiting takeover run to an existing same-bot user lease. In a shared Team Computer, however, a different bot can still own the active user lease. Binding the requesting bot's run to that foreign lease breaks the release invariant: releasing the owner bot clears controlRunId, but cannot resume a run owned by the requester, leaving it in waiting_takeover.

## What

- Preserve an active foreign user lease without attaching the requesting bot's run to it.
- Let the existing takeover flow revoke the foreign lease and bind the waiting run to the new same-bot lease.
- Tighten the transaction test to assert that foreign lease ownership never receives another bot's run binding.

## Tests

- pnpm --filter @rakazo/db test
- pnpm --filter @rakazo/db check
- pnpm exec biome check packages/db/src/events.ts packages/db/src/events.test.ts

Follow-up to #658.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved takeover handling when a computer has an active user lease owned by another bot.
  - Preserved the existing lease without incorrectly assigning control to the current run.
  - Added test coverage to verify that foreign active leases remain unchanged during takeover.

### DIFF
--- a/packages/db/src/events.test.ts
+++ b/packages/db/src/events.test.ts
@@ -580,7 +580,7 @@ describe("pauseRunForTakeover", () => {
     );
   });
 
-  it("preserves another bot's active user lease and only binds controlRunId", async () => {
+  it("preserves another bot's active user lease without binding this run to it", async () => {
     const expiresAt = new Date(Date.now() + 60_000);
     const tx = {
       $queryRaw: vi.fn().mockResolvedValue([{ id: "computer-1" }]),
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Looks good. Same-bot retain still binds `controlRunId`; foreign active lease stays untouched until takeover revokes it. Test matches. CI green.

PR #669, opened 2026-09-05
fix(computer): keep takeover leases with their winner

fetw882:
## Why

Follow-up to the real shared-Team-Computer failure reported in #658. Two replacement takeovers can race after both observe the same previous user lease. One request wins the compare-and-set; the loser currently returns the winner's active lease ID, even though `controlBotId` belongs to the other bot. The losing screen then appears to have control but cannot use or release that lease.

## What

- Return an idempotent active lease only when it is owned by the requested bot.
- Return `CONFLICT` when another bot won the takeover race, without exposing that bot's lease ID.
- Add a deterministic three-bot product journey that barriers both revocations, races the two replacement takeovers, and verifies exactly one success, one conflict, and a response lease matching the persisted winner.

## Tests

- `pnpm exec biome check apps/api/src/router.ts packages/testkit/src/journeys.test.ts`
- `pnpm --filter @rakazo/api check`
- `pnpm --filter @rakazo/testkit check`
- `git diff --check`

The Postgres journey is compiled locally and will execute in the repository Postgres journeys CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent computer takeovers to prevent control from being assigned to the wrong bot.
  * Conflicting takeover attempts now return a clear conflict error instead of proceeding silently.
  * Ensured successful takeovers are linked to the correct control lease.

### DIFF
--- a/apps/api/src/router.ts
+++ b/apps/api/src/router.ts
@@ -1519,17 +1519,17 @@ export function createRouter(deps: RouterDeps) {
           const current = await deps.prisma.computer.findUniqueOrThrow({
             where: { id: bot.computer.id },
           });
-          if (!hasActiveComputerControl(current)) throw new ORPCError("CONFLICT");
-          if (current.controlBotId === bot.id) {
-            await bindWaitingTakeoverToControl(deps, {
-              spaceId: context.actor.spaceId,
-              threadId: bot.thread?.id,
-              botId: bot.id,
-              computerId: current.id,
-              controlLeaseId: current.controlLeaseId!,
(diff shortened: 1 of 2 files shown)


fetw882:
Addressed the accepted Greptile test-harness concern on the new head: the concurrent-revocation barrier now has a one-second failure fuse and is unconditionally released/cleared in `finally`, so an early request failure reports promptly instead of waiting for the journey timeout. Changed-file Biome, testkit typecheck, and diff check pass.

PR #675, opened 2026-09-05
fix(adapters): bound Docker supervisor responses

fetw882:
## Why

Successful responses from the Docker sandbox supervisor were still parsed with unbounded `Response.json()` calls. A compromised, buggy, or mismatched supervisor could therefore make the API/worker buffer an arbitrary response for provision, command execution, screen operations, observations, actions, file listing, or file reads.

This follows the repository feedback to batch same-theme response-bound changes behind one shared helper instead of opening one PR per endpoint.

## What

- add one streaming `readSandboxJson` helper with a 16 MiB success-response cap and a 30-second body-read deadline
- reject oversized declared bodies before buffering and stop streamed bodies as soon as they cross the cap
- route all seven Docker supervisor JSON response call sites through the helper
- derive a tighter base64-envelope cap when `readFile` already has a caller-provided byte limit
- keep the existing small, short-deadline error-body path unchanged

No user-facing copy is changed. The new error text is operator-facing and includes only the enforced byte limit.

## Test

- `pnpm exec vitest run packages/adapters/src/docker-sandbox.test.ts` (11 passed)
- `pnpm --filter @rakazo/adapters test` (1,134 passed, 7 skipped)
- `pnpm --filter @rakazo/adapters check`
- `pnpm exec biome check packages/adapters/src/docker-sandbox.ts packages/adapters/src/docker-sandbox.test.ts`
- `git diff --check`

Regression coverage verifies both an oversized declared response (including body cancellation) and a streamed file response crossing the caller-derived encoded limit.

### DIFF
--- a/packages/adapters/src/docker-sandbox.test.ts
+++ b/packages/adapters/src/docker-sandbox.test.ts
@@ -88,7 +88,7 @@ describe("Docker sandbox", () => {
           new Response(
             new ReadableStream({
               start(controller) {
-                controller.enqueue(new Uint8Array(1027));
+                controller.enqueue(new Uint8Array(1029));
               },
               cancel,
             }),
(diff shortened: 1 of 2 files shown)


fetw882:
Addressed the Greptile file-read contract edge in 637286bf: explicit maxBytes now determines the base64-plus-JSON response bound, while operations without a caller limit keep the fixed 16 MiB cap. Added a regression above the generic cap. Focused tests: 12 passed; adapters rerun: 1135 passed, 7 skipped; typecheck and Biome pass.

PR #474, opened 2026-09-01
feat(adapters): GraphQL as optional tool source

elie222:
## Why

Agents should be able to use an existing GraphQL API without a custom connector or hosted gateway. Installing an HTTPS endpoint now discovers query and mutation tools through the same optional connector catalog used by MCP and OpenAPI.

## What changed

- Add endpoint introspection, bounded selections including union results, authenticated execution, and shared contracts for GraphQL tool sources.
- Keep credentials encrypted and reuse the existing SSRF-safe transport, header restrictions, and bounded response reader.
- Reject HTTP, GraphQL, malformed JSON, invalid envelopes, and oversized validation responses as tool errors; preserve nullable result fields and defaulted root arguments.
- Add installation to web/Electron and mobile Advanced integrations, retaining current components and localization.
- Merge current main, retain the fixes from #677 and #714, consolidate duplicated connector HTTP helpers, and make selection-budget accounting linear.

## UI

The Advanced action is “Add GraphQL”; its form is “Add GraphQL endpoint” on web and “GraphQL endpoint” on mobile, with the placeholder `https://example.com/graphql`. The empty state is “No tool sources yet.” These labels identify the new protocol and its endpoint only when Advanced is open; the empty state appears only when no source is installed. The default source name is “GraphQL” and its compact source marker is “G”.

The dedicated web E2E covers Advanced ordering and installation. Native mobile uses the same install API; web screenshots do not claim native mobile coverage.

CI screenshots: [GraphQL installation](https://rakazogithubactions.fsn1.your-objectstorage.com/playwright/runs/34023891209-1/screenshots/images/013-02-graphql-installed.png), [Advanced actions](https://rakazogithubactions.fsn1.your-objectstorage.com/playwright/runs/34023891209-1/screenshots/images/002-01-graphql-advanced-order.png). These frames are from final commit d04201b2.

## Validation

- All 22 workspace check tasks pass.
- Repository lint passes with existing warnings.
- Adapter suite: 1,402 tests passed and 25 skipped; one local Docker conformance test failed because the supervisor rejected its home-directory setting. The failing path is unchanged by this PR.
- GraphQL regressions cover unsafe auth headers, defaulted non-null arguments, malformed envelopes, rejected top-level null data, valid nullable fields, protocol and HTTP errors, union selections, and optional-off discovery.
[...]

Issue #414, opened 2026-08-30
GraphQL as a user-installed tool source


elie222:
MCP, OpenAPI, and Treg already exist. Add GraphQL the same way: paste an endpoint, introspect, expose operations as tools. Optional. No hosted vendor.

PR #677, opened 2026-09-05
fix(graphql): select union result fields

fetw882:
## Why

Issue #414 asks for user-installed GraphQL endpoints whose operations behave as tools. PR #474 implements that path, but an operation returning a union currently selects only __typename because union definitions have no fields of their own. The resulting tool succeeds while dropping every useful member field.

## What

- include possibleTypes in the bounded introspection document
- generate inline fragments for concrete union members
- enforce the existing selection-depth and 6000-character budget at every nesting level
- support unions both as root results and as nested fields
- cover the generated persisted operation and its execution document

No user-facing copy is added or changed.

## Test

- pnpm exec vitest run packages/adapters/src/graphql-connectors.test.ts (13 passed)
- pnpm --filter @rakazo/adapters test (988 passed, 7 skipped)
- pnpm --filter @rakazo/adapters check
- pnpm check (20 workspaces passed)
- pnpm exec biome check packages/adapters/src/graphql-connectors.ts packages/adapters/src/graphql-connectors.test.ts
- git diff --check

This stacked branch also carries the Expo SDK patch-version alignment already merged to main so the older target branch passes its current compatibility check; no unrelated mobile or web behavior is changed.


fetw882:
Addressed the nested-union budget finding in 92fd90ef. Every scalar, nested field, and union fragment now reserves space before append, so every recursion level stays within the 6000-character persisted-selection budget. Added a broad-schema regression. Focused tests: 13 passed; adapters: 988 passed, 7 skipped; adapter typecheck and Biome pass. The current full-repo Typecheck failure is the target branch baseline: Expo install --check expects the SDK patch versions already present on main via #467; it is unrelated to these two adapter files.

fetw882:
The target-branch Typecheck drift is now removed in 628f9340: I carried forward the two Expo patch alignments already present on main, while preserving the target branch's unrelated web UI. Local pnpm check now passes all 20 workspaces; the 13 GraphQL tests and Biome check remain green.

PR #678, opened 2026-09-05
fix(supervisor): reject computer control redirects

elie222:
## Why

A compromised Docker computer can replace its desktop control listener and redirect the next authorized observation or action. The supervisor currently follows that response using its own network access, escaping the initially inspected computer and fixed desktop endpoint. In the isolated Compose topology, the supervisor joins every computer network and also has access to its own loopback and application network.

The issue is present on current main. The listener and bot commands share the container user; dropping capabilities and forbidding privilege escalation do not prevent same-user listener replacement. The shell lifecycle filter blocks common command spellings but is not an operating-system boundary for arbitrary programs. Supervisor bearer authentication and bot/Space label checks authorize the initial request, not its redirect destination.

Assessed severity: **Medium**, with deployment-dependent higher impact if sensitive unauthenticated HTTP services are reachable. This demonstrates constrained server-side request forgery, not host takeover: other computers require different control tokens, supervisor operations require a separate service credential and identity headers, cross-origin fetch redirects remove Authorization, and HTTP redirects cannot access the Docker Unix socket. Responses must pass the control JSON/completion checks; ordinary bodies are not returned intact, though error details can surface through actions. The default remote-provider production topology does not use this Docker control path; host-run Docker Desktop setups may already fall back when container IPs are unreachable.

Checked the existing loopback-control, Team desktop, resource-limit, and supervisor-response-bounds proposals; none rejects these redirects.

## What changed

Reject redirects in the desktop control fetch, keeping requests on the Docker-inspected address, control port, and fixed endpoint. Redirect rejection remains an action failure, preventing unsafe replay. Observations retain their existing Docker-exec fallback on the original computer. No provider contracts or UI copy change.

## Validation

- Deterministic local HTTP tests with fake tokens and responses cover 301, 302, 303, 307, and 308, both relative endpoint changes and a second origin. Each verifies zero requests to the redirect target, no action replay, and observation fallback. Direct requests retain their method, token, mapped actions, and response.
- Before the fix, all ten redirect cases failed by contacting the forbidden target. The reproduction observed GET for 301/302/303, POST for 307/308, and removal of Authorization across origins.
- After the fix, all twelve regression tests pass. The supervisor suite passes 84 tests, with one existing Linux-only test skipped. Supervisor typechecking and repository lint pass; lint reports existing warnings outside this change.
- No production services or local Electron E2E were used. Docker network reachability was traced from the checked-in topology; the HTTP reproduction uses loopback servers rather than live containers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Desktop control requests no longer follow redirects to unintended endpoints.
  - Desktop actions are not replayed after a redirect, helping prevent duplicate interactions.
  - Observation now falls back to the original computer when a redirect occurs.

- **Tests**
  - Added coverage for direct desktop control responses, request parameters, redirect handling, and observation fallback behavior.

### DIFF
--- a/infra/sandboxes/supervisor/src/computer-control.test.ts
+++ b/infra/sandboxes/supervisor/src/computer-control.test.ts
@@ -1,0 +1,116 @@
+import http from "node:http";
+import { afterEach, describe, expect, it } from "vitest";
+import { controlDesktop } from "./index.js";
+import {
+  attemptComputerControl,
+  preferComputerControl,
+  shouldReplayComputerActions,
+} from "./supervisor-logic.js";
+
+const token = "test-computer-control-token";
+const actions = [{ kind: "key" as const, key: "Enter" }];
+const observation = { image: "ZmFrZQ==", mimeType: "image/png", width: 1, height: 1 };
+const servers: http.Server[] = [];
+
+async function listen(handler: http.RequestListener) {
+  const server = http.createServer(handler);
+  servers.push(server);
+  await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+  const address = server.address();
+  if (!address || typeof address === "string") throw new Error("expected a TCP address");
+  return `http://127.0.0.1:${address.port}`;
+}
+
+afterEach(async () => {
+  await Promise.all(
+    servers.splice(0).map(
+      (server) =>
+        new Promise<void>((resolve, reject) => {
+          server.close((error) => (error ? reject(error) : resolve()));
+          server.closeAllConnections();
+        }),
+    ),
+  );
+});
+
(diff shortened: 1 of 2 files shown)


elie222:
Validation is complete: lint, typechecking, production builds and the CI Electron smoke, unit tests, Postgres journeys, and Web E2E all pass. Both automated code reviews finished without actionable code findings, and there are no unresolved review threads.

The generic docstring-coverage advisory is non-blocking. The redirect boundary is documented beside the fetch and the regression cases describe their behavior. Keeping the existing function documentation style preserves the scope of this security fix.

The PR is left unmerged for maintainer review.

PR #681, opened 2026-09-05
fix(ci): isolate untrusted Playwright artifacts

elie222:
PR screenshot publishing must treat contributor artifacts as data. Extracting an artifact into the trusted checkout allowed it to replace the publisher, comment generator, or their configuration before those files ran with storage credentials or a token that can write issues and pull requests.

Download the already-validated artifact by ID and extract it into a fresh runner temporary directory. A small standard-library extractor validates every ZIP entry before writing, rejecting absolute paths, parent traversal, backslashes, links, and special files. Pass report and screenshot directories explicitly to the existing trusted publisher; keep generated review data in the trusted workspace. This closes both the publisher and comment-generator paths in one change.

Validation: 46 focused tests passed, including 10 offline regressions for replacement scripts/configuration/dependencies/review data, unsafe ZIP entries, existing destination links, PR galleries and comments, main reports, and missing-artifact timeout galleries. Publication tests run the real scripts with fake credentials and local storage. Repository lint passed with existing warnings. No production services or local Electron E2E were used. No overlapping open or merged fix was found before opening this PR.

Final validation: CI passed lint, typecheck, the full unit suite, Postgres journeys, production builds including the CI-only Electron smoke tests, web and marketing E2E, and all image validation checks. Greptile reviewed the final revision without blocking findings. CodeRabbit reviewed the initial fix; its fixture-maintainability feedback and documentation warning were addressed, but its follow-up review was rate-limited. No unresolved review threads remain.

### DIFF
--- a/.github/workflows/publish-playwright-report.yml
+++ b/.github/workflows/publish-playwright-report.yml
@@ -91,6 +91,7 @@ jobs:
             <<<"$artifacts")"
           if [[ "$count" == "1" ]]; then
             echo "exists=true" >> "$GITHUB_OUTPUT"
+            echo "id=$(jq -r --arg name "$ARTIFACT_NAME" '.artifacts[] | select(.name == $name and .expired == false) | .id' <<<"$artifacts")" >> "$GITHUB_OUTPUT"
           elif [[ "$count" == "0" && "$PLAYWRIGHT_JOB_RESULT" == "timed_out" ]]; then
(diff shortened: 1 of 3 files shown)


elie222:
Addressed the fixture-mode feedback: ZIP modes are now keyed explicitly by entry name, so adding or reordering fixture entries cannot change which entry is a symlink. Added concise helper documentation for the coverage warning. All 46 focused tests passed again, including the 10 publication and extraction regressions.

PR #689, opened 2026-09-05
fix(memory): keep concurrent saves consistent with revision history

elie222:
Concurrent saves to the same Markdown memory document could overwrite its current content before colliding on a revision number. A save could report failure after changing memory, leaving the current document inconsistent with its history. This affects authenticated memory edits and agent memory writes through the shared backend.

The document lookup, write, and revision insert now run in one serializable transaction. The existing database retry helper retries transient conflicts from fresh state, with a limit of three attempts. Each committed replacement keeps its matching history and provenance; rejected transactions leave no partial write. Other database errors are returned without automatic retries.

Validation:
- Seven focused memory tests pass, including five deterministic PostgreSQL regressions using an isolated database and synthetic data.
- The five new regressions fail against the original implementation. They cover concurrent user/bot saves, rejected history inserts on create/update, exhausted conflict retries, and successful caller retries.
- Memory package typechecking and lint checks for the changed files pass.
- The PostgreSQL regressions are included in the existing integration CI job.

This change prevents new inconsistencies; it does not repair historical gaps or change the nullable document-key schema. No UI changes.

### DIFF
--- a/packages/memory/src/commit.postgres.test.ts
+++ b/packages/memory/src/commit.postgres.test.ts
@@ -1,0 +1,238 @@
+import type { AdapterContext, MemoryCommitRequest } from "@rakazo/adapter-kit";
+import { createDb, type PrismaClient } from "@rakazo/db";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { MarkdownMemoryStore } from "./index.js";
+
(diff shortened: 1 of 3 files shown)


elie222:
@coderabbitai review

PR #690, opened 2026-09-05
fix(updater): restore fork checkout before Compose recovery

elie222:
## Why

When a fork update changed Compose configuration and then failed to recreate healthy services, automatic recovery reused the previous image tag while still reading files from the failed revision. Restoring the checkout afterward could leave the server running the wrong service configuration or unable to recover.

This affects self-hosted installations using the optional updater sidecar to build a fork. The API requires the deployment owner and the sidecar requires its bearer token. An attacker would need a chosen fork revision to be deployed and fail; this is a low-severity security/reliability issue with availability and configuration impact, without a new authentication or privilege bypass. Published-image updates do not change the checkout.

## What changed

- Restore the original branch or detached commit before restoring the image pin and invoking Compose recovery. Refuse fork updates when the original commit cannot be verified.
- Attempt checkout restoration even when checkout or merge fails partway through. Skip Compose and require manual recovery if the original checkout cannot be restored, retaining the primary failure and remote cleanup.
- Reuse the existing configured Compose target. Open PR #619 adds overlays and extra services but does not fix this ordering; this change leaves its configuration and command-building work intact.
- Exercise the production command runner with fake Git and Docker executables and temporary deployment files, including changed Compose and referenced configuration, custom Compose paths, detached checkouts, restoration failures, and successful updates.

## Validation

- All 127 focused updater and shared Compose tests passed with one worker.
- Updater TypeScript checking and Biome checks for the changed files passed.
- Negative control: six of eight new tests failed against the original implementation, including observing the old image tag combined with the failed revision's configuration. The fixed implementation passes all eight.
- No real updater, Docker deployment, production service, or local Electron E2E was run. The fake-executable tests cover orchestration and file state, not Docker's reconciliation internals; restoring files and images also cannot undo database migrations or other persistent side effects.

## Conditional error copy

- "The deployment checkout's current commit could not be verified. Fix it before updating."
- The recovery advice adds "was skipped because the previous checkout could not be restored" when checkout restoration fails.

Both messages appear only after the corresponding failure. They explain why the update was refused or why manual recovery is necessary; removing the cause would leave the operator unable to distinguish a checkout problem from an image health failure. No persistent explanatory copy is added.

### DIFF
--- a/infra/updater/src/fork-recovery.test.ts
+++ b/infra/updater/src/fork-recovery.test.ts
@@ -1,0 +1,297 @@
+import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import type { ServerUpdateRun } from "@rakazo/contracts";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createUpdaterApp } from "./index.js";
+import { DEFAULT_COMPOSE_FILE, resolveUpdaterConfig } from "./updater-logic.js";
+
+const token = "fake-updater-recovery-token-0000000000";
+const currentCommit = "1".repeat(40);
+const targetCommit = "2".repeat(40);
+const oldTag = `local-${currentCommit}`;
+const originalRemote = "https://github.com/example/previous-fork";
+const nextRemote = "https://github.com/example/next-fork";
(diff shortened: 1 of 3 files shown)


elie222:
@coderabbitai review

PR #692, opened 2026-09-05
fix(adapters): settle runtime work on stream cancellation

elie222:
## Why

Stopping consumption of a Pi run could leave its model and tool work executing after the worker released its screen and computer lease. A supplied context signal also replaced the runtime's own cancellation signal, making `runtime.abort()` ineffective in normal executor use. This affects already-running Pi-backed bots across clients sharing the backend. The realistic impact is continued resource use or tool effects after cancellation; it is not an unauthenticated entry point.

Validated against current main, including the earlier sandbox-process teardown and cancellation-finalization fixes. Those changes do not settle the detached Pi agent loop. Three deterministic regression cases fail against the original runtime because cancellation never reaches the fake agent.

## What changed

- Compose caller and runtime cancellation, abort immediately when an iterator is closed (including a pending event read), and await the agent work before dropping run tracking.
- Await nested-agent shutdown and remove abort listeners even when prompting fails. Keep a replacement attempt's tracking when an older stream closes.
- Cancel executor tool work before closing an interrupted runtime iterator, reject new tool calls after cancellation, and stop post-stream processing after lease loss. Normal completion keeps its context available for checkpoints; existing pause/resume behavior remains intact.

This fixes worker/runtime cleanup. Cross-process Stop detection still uses the existing event and lease checks, including the heartbeat for quiet streams. Cancellation cannot undo an external effect already accepted by a provider; work that ignores cancellation must settle before this worker finishes cleanup.

## Validation

- All 106 focused tests passed with one worker, covering Pi runtime behavior, cancellation ordering, executor behavior, approval replay, protected-input pauses, and scripted runs. All model work in the new tests is fake and offline.
- Adapter TypeScript checks, targeted Biome checks, and whitespace checks passed.
- CI passed: repository lint and typechecks, unit tests, Postgres journeys, production builds with Electron smoke tests, Web E2E, and container image validation. No UI changes.

### DIFF
--- a/packages/adapters/src/executor.ts
+++ b/packages/adapters/src/executor.ts
@@ -209,6 +209,7 @@ import {
   secretPausedToolResult,
   tryCompleteConnectionWithCode,
 } from "./run-secret.js";
+import { withRuntimeCleanup } from "./runtime-stream.js";
 import {
   cancelScheduleFromTool,
   createScheduleFromTool,
@@ -1390,6 +1391,7 @@ export function createRunExecutor(deps: ExecutorDeps) {
           args: Record<string, unknown>,
           executionId: string,
         ) => {
+          context.signal.throwIfAborted();
           if (handedOff) {
             return { error: "This stage was handed off. End the turn without more tool calls." };
           }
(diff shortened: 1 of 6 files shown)


elie222:
@coderabbitai review

PR #695, opened 2026-09-05
fix(adapters): enforce approvals for read-only connector tools

elie222:
## Why

A connector tool that advertises `readOnly: true` can run despite a matching require-approval rule. Remote MCP `readOnlyHint` flows into this metadata. The executor resolves the approval policy correctly, but skips the effect record that enforces approval and automatic review.

This is a conditional, medium-severity approval bypass: a configured connector must advertise the hint and its tool must be invoked. Impact depends on the connected tool's permissions. The shared executor affects web, desktop, and mobile. Current main and existing approval PRs were checked before implementing this fix.

## What changed

Record connector calls through the existing effect lifecycle regardless of read-only metadata. Explicit approval rules, persisted denials, exact approved replay, and automatic-review decisions now take effect on both direct and lazy-catalog calls. Resumed approvals retain their stable effect key when choosing Always allow changes the rule, so the approved record is consumed rather than left available for another replay. Existing rule precedence and default read behavior are preserved: ordinary reads run without prompts and separate read calls still execute independently.

The operational tradeoff is additional effect records for connector reads. Built-in tool policy and run cancellation are unchanged.

## Validation

- Reproduced the bypass on unchanged main with offline executor tests: explicit approval and automatic-review cases failed while ordinary reads passed.
- All 93 focused approval, replay, and automatic-review tests pass, including 20 new executor regressions covering direct and catalog routes.
- Adapter typecheck and focused Biome checks pass.
- No live services or production data were used. Tests ran with one worker. Full-suite and Electron verification are left to CI.

### DIFF
--- a/packages/adapters/src/executor-readonly-approval.test.ts
+++ b/packages/adapters/src/executor-readonly-approval.test.ts
@@ -1,0 +1,317 @@
+import type { AgentRunRequest, ConnectorCall, ConnectorTool } from "@rakazo/adapter-kit";
+import type { ActionApprovalRule } from "@rakazo/core";
+import { approvalEffectKey } from "@rakazo/core/node/approval-effect-key";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { isApprovalPausedResult } from "./approval-effect.js";
+import { runAutoReviewJudge } from "./auto-review.js";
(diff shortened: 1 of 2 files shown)


elie222:
@coderabbitai review

PR #696, opened 2026-09-05
fix(core): disable unsafe chart links

elie222:
Charts can carry links supplied through bot-generated specs or external data. Previously, the shared renderer preserved executable SVG links, so activating a crafted mark could run JavaScript in the signed-in app context. This is interaction-dependent stored XSS: an attacker must influence a published chart and induce link activation. The live web UI and Electron's shared web UI use this renderer; mobile currently displays a text placeholder, and PNG exports are inert.

Filter resolved anchor URLs in the shared plot renderer before mounting or serialization. Allow HTTP, HTTPS, mail, telephone, and relative links using browser-compatible URL parsing; remove unsafe or malformed href attributes while preserving chart marks and caller-owned data. Filtering the output covers data columns, per-mark arrays, channel objects, scales, transforms, and previously saved specs. No product UI copy is added.

Validation:
- Confirmed JavaScript execution in headless Chromium with the original renderer, including mixed-case/control-character schemes and percent-encoded script text; the payload only set a test marker.
- All 41 focused plot tests pass, covering live output, serialized SVG, legitimate links, channel variants, existing chart examples, and PNG rendering.
- Both focused Chromium tests pass through the real chart component, including link activation with hover inspection enabled and disabled, expanded charts, hover redraws, and intercepted legitimate/encoded-relative navigation.
- Core and adapters type checking, changed-file lint, and diff whitespace checks pass. Local tests use a single worker; CI runs the broader suites and Electron E2E.

[CI screenshot: chart with its unsafe link disabled](https://rakazogithubactions.fsn1.your-objectstorage.com/playwright/runs/33991962382-1/screenshots/images/099-chart-with-inert-link.png).

All CI checks pass on the final commit, including web and Electron E2E. Greptile reviewed the final commit without actionable findings. CodeRabbit could not review because its quota was exhausted; a retry after the initially reported cooldown was also rate-limited. The PR remains unmerged with auto-merge disabled.

### DIFF
--- a/apps/web/e2e/chart-links.spec.ts
+++ b/apps/web/e2e/chart-links.spec.ts
@@ -1,0 +1,81 @@
+import { expect, test } from "@playwright/test";
+import { captureScreenshot } from "./helpers";
+
+const fixture = "/e2e/fixtures/chart-links.html";
+
+test("chart links cannot execute script when clicked or expanded", async ({ page }, testInfo) => {
+  for (const inspect of ["0", "1"]) {
+    for (const href of [
+      'javascript:void(document.body.dataset.plotExecuted="yes")',
+      ' \u0000\u001fJaVa\tScRi\nPt:void(document.body.dataset.plotExecuted="yes")',
+      'javascript:%76oid(document.body.dataset.plotExecuted="yes")',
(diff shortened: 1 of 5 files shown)


elie222:
@coderabbitai review

PR #698, opened 2026-09-05
fix(screen): sandbox proxied documents on direct navigation

elie222:
Screen HTML served behind a signed capability URL inherits the application origin when opened directly, bypassing the iframe sandbox. A controlled screen listener and a signed-in user opening a valid capability URL can therefore expose app-origin data and authenticated API responses. This is a conditional high-impact issue; changing desktop pixels alone is insufficient, and Docker already mounts bundled noVNC files read-only.

Add a shared response policy in `@rakazo/core` that enforces `sandbox allow-scripts allow-pointer-lock` as a separate CSP policy on every proxied HTTP response. Existing upstream CSP and framing restrictions remain intact, and upstream policy cannot restore same-origin access. Keep module CORS access, response cookie filtering, and WebSocket negotiation intact.

Production Compose serves screen content through Vite preview; the API issues capabilities using the configured web origin and does not have a separate screen response route. The same policy covers development, production preview, and proxied URLs loaded by web, Electron, and mobile. The browser regressions exercise the real Vite configuration with API-issued capabilities and local fake listeners.

Validation:

- Seventeen focused unit tests passed for capability issuance, authorization, response isolation, and handshake filtering.
- Six headless Chromium regressions passed with one worker, covering direct navigation and frames with and without the iframe sandbox in development and preview. Module imports, WebSocket traffic, and configured framing remain usable while cookie, storage, parent-DOM, and app-response reads are blocked.
- Removing only the response sandbox made both direct-navigation tests fail by exposing fake cookie, storage, and authenticated response data; restoring it passed the suite.
- Core and web typechecks, a standalone typecheck of the new browser fixtures, and lint of changed files passed.
- An additional standalone check of the Vite configuration reports existing `protocol` property typing errors in target resolution; the configuration is outside the normal web typecheck. Provider-backed and native rendering were not exercised locally; Electron E2E remains with CI.

### DIFF
--- a/apps/web/e2e/screen-proxy-isolation.spec.ts
+++ b/apps/web/e2e/screen-proxy-isolation.spec.ts
@@ -1,0 +1,183 @@
+import { createHash } from "node:crypto";
+import { mkdir, mkdtemp, rm } from "node:fs/promises";
+import { createServer as createHttpServer, type Server } from "node:http";
+import type { AddressInfo, Server as NetServer } from "node:net";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { expect, test } from "@playwright/test";
+import { createServer, type Plugin, preview, type ViteDevServer } from "vite";
+import { addScreenProxyCapability } from "../../api/src/screen-proxy";
+
+const secret = "fake-screen-proxy-browser-test-secret";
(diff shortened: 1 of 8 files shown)


elie222:
@coderabbitai review

elie222:
@coderabbitai full review

Issue #719, opened 2026-09-06
Per-bot model choice is buried in Advanced and easy to miss


RA-h:
### Confirmation

- [x] This is not a security vulnerability report

### Description

 A per-bot model override exists, but it lives inside the collapsed Advanced section of bot settings. On first glance the panel shows only name, title, and description, so users reasonably conclude one default model applies to all bots and never discover the override.

### Steps to reproduce

1. `pnpm dev`, create an account, connect a model, and create two bots
2. Open the second bot's settings panel
3. Observe no model choice is visible until the Advanced section is expanded

### Expected behavior

 The per-bot model choice (keeping the "Space default" fallback option) should be visible at the top level of bot settings, not hidden behind the collapsed Advanced section. Mobile bot settings should offer the same choice or have an explicit reason for omitting it.

### Actual behavior

 Model (and Thinking level) are hidden inside collapsed Advanced on web, and there is no model choice in mobile bot settings at all — so the override is effectively undiscoverable.

### Rakazo version or commit

main @ cb804af9

### Operating system

Windows

### Logs or screenshots

```shell

```

elie222:
Closing aligned with https://github.com/elie222/rakazo/pull/723: per-bot model stays under Advanced. Create-time model is #713 / #720.

elie222:
Also closing the follow-up PR https://github.com/elie222/rakazo/pull/730 without merge.

Product call stands: the per-bot model override stays in **Advanced** on web (and we are not promoting it into the main settings chrome). Mobile still needs a discoverable Advanced path for the same control, but that is separate from pulling Model out of Advanced on web.

Create-time / spawn model selection is tracked under #720 (and the create-flow work), not by moving the settings control.

PR #733, opened 2026-09-06
feat(mobile): add Advanced model controls

fetw882:
## Why

Mobile bot settings did not expose the per-bot model and thinking overrides already available on web. A collapsed Advanced section makes those controls available without adding them to the main settings surface. Web remains unchanged.

Follow-up to #719 and https://github.com/elie222/rakazo/issues/719#issuecomment-5560345633.

## What

- Add a collapsed Advanced disclosure to mobile bot settings.
- Expose the connected-model picker and conditional thinking-level picker inside it, using the shared native action-sheet helper and its Android pagination.
- Persist the existing paired modelProvider/modelId override. Omit unchanged model metadata so unrelated saves cannot overwrite a concurrent update.
- Preserve thinking overrides unless the model or thinking control changes.
- Add Simplified Chinese strings and Maestro assertions for the Advanced path.

## Visible copy

- **Advanced** names the collapsed entry point; hiding the entry itself would make the controls unreachable.
- **Model** and **Space default** identify the override and inheritance choice. They appear only after expanding Advanced.
- **Thinking**, **Default (medium)**, **Minimal**, **Low**, **Medium**, **High**, **Extra high**, and **Max** label the supported reasoning choices. The control appears only when the effective model supports it, and option labels appear in its picker.
- **Could not load model settings** appears only if model metadata loading fails, so unavailable metadata does not look like an empty model list.
- Existing **Cancel** and **More** labels are reused by the native picker; no persistent explanatory copy is added.

## Test

- Rechecked the current PR head: mobile unit tests passed, 27 files / 190 tests; mobile TypeScript check passed.
- Original implementation verification: full workspace check passed (22 tasks) and Biome passed on changed TypeScript files.
- The latest existing PR CI and automated reviews passed. No-conflict integration with the reviewed main snapshot was checked separately; this is not a substitute for running the merged tree.

## Native screenshots

Native screenshots have not yet been captured for this PR head. The Maestro flow includes Advanced/model assertions, but those assertions are not themselves visual evidence. Regular PR CI does not run the dedicated manual Android screenshot workflow; that workflow is present in the repository but is not currently registered in the fork's workflow list. No unrelated web screenshot is offered as evidence for this native-only change.

## Summary by CodeRabbit

- **New Features**
  - Added advanced bot settings for choosing the model and thinking level.
  - Added a collapsible Advanced section with Model and Space default options.
  - Settings now load and save model and thinking-level preferences.
  - Added Chinese translations for the new settings and status labels.

- **Tests**
  - Updated the mobile screenshot flow to verify the Advanced settings and available options.


fetw882:
Addressed both review findings in 8d8667bf: saves now omit `thinkingLevel` whenever model metadata is unavailable, preserving the stored override; Android model/thinking choices now use the repository’s existing paginated native `Alert` action-sheet helper instead of a custom modal. Re-ran mobile typecheck and all 190 mobile tests.

PR #737, opened 2026-09-06
test(compose): run installer smokes in CI

fetw882:
## Why

Compose installer smoke scripts cover offline startup, proxy precedence, and mirrored-image startup documentation, but they currently run only when invoked locally. Regressions in these shell-only paths can therefore pass pull-request CI.

## What changed

- add a lightweight `Compose installer smokes` job to the compose/image workflow
- run the existing aggregate smoke runner on relevant pull requests
- keep the job independent of Node and Docker setup, with read-only permissions and a five-minute timeout

Follow-up to #534.

## Tests

- `pnpm test:compose-smoke` (all four smoke scripts passed)
- workflow YAML parsed successfully
- `git diff --check`

## Summary by CodeRabbit

* **Tests**
  * Added automated smoke testing for the Compose installer on pull requests.
  * Installer checks now run with a five-minute timeout and read-only repository access.

### DIFF
--- a/.github/workflows/publish-server-image.yml
+++ b/.github/workflows/publish-server-image.yml
@@ -33,6 +33,19 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  installer-smoke:
+    name: Compose installer smokes
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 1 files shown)


elie222 (approved):
Looks good — lightweight PR-only job that runs the existing bash compose installer smokes (no Node/Docker), pinned checkout, read-only perms, 5m timeout. Matches the rest of this workflow.

Issue #401, opened 2026-08-30
Routine event triggers (webhook, GitHub, Slack)


elie222:
Cron already exists. Add first-class event triggers: plus on Routines, then Add trigger.

Webhook, GitHub, and Slack-style "when this happens, wake." Related closed work: #195 / #334 (trigger UX).

PR #740, opened 2026-09-06
feat(routines): add signed GitHub event triggers

fetw882:
## Why

Routines expose a disabled Git event option, but repository events currently need custom glue through the generic bearer-token webhook. That path does not verify GitHub delivery signatures or preserve GitHub delivery ids for deduplication. This adds the first complete provider event slice from #401 without introducing a hosted-service dependency.

Progresses #401.

## What changed

- add a persisted `githubEnabled` routine trigger across Prisma, contracts, and routine RPCs
- accept GitHub deliveries at `/api/v1/bots/:botId/github` only after validating `X-Hub-Signature-256` with constant-time HMAC-SHA256 comparison
- keep the existing 64 KiB request bound, reject oversized bodies before target lookup, select only active GitHub-enabled routines, and acknowledge valid deliveries without waking a bot when no routine matches
- hash `X-GitHub-Delivery` into the run nonce so retries do not create duplicate work
- exclude event-authored strings from the agent prompt; only allowlisted event/action enums, numeric IDs, booleans, and commit hashes can cross the GitHub wake-up boundary
- fence structured, plain-text, and malformed-JSON generic webhook bodies as untrusted delivery data instead of promoting external text to agent instructions
- force webhook-triggered runs to request owner approval before local or connector side effects, including `shell` and `write_file`, even when an always-allow rule exists; read-only inspection remains unattended
- coalesce concurrent per-bot secret provisioning so trigger setup and Save cannot rotate the same secret twice
- expose Git event setup in the web/Electron routine editor and show the trigger on the mobile routine detail surface
- add a CI E2E path that saves the trigger and captures its configuration screen

The GitHub Secret field reuses the existing encrypted per-bot inbound secret and rotation path.

## UI copy

- `Git event` identifies the new trigger in the picker, configured card, routine summary, and mobile detail. It appears only when selecting or viewing the trigger, so removing it would make the event source ambiguous.
- `X-Hub-Signature-256: sha256=…` is the exact technical header users need to recognize when configuring GitHub and is shown only inside the selected trigger card.
- `Add a schedule, webhook, or GitHub trigger to run this routine.` updates the existing empty-state validation to include the newly valid trigger.
- `Remove Git event` is an accessibility label for the icon-only remove control.

## Verification

- 83 focused webhook, emulator, and approval-boundary tests pass, including plain-text/malformed-JSON framing and forced `shell`/`write_file` approval regressions
- web unit suite passes (204 tests), including concurrent secret-provisioning coverage
- all 22 workspace typecheck tasks pass
- web production build passes
- Biome and whitespace checks pass on the changed files
- desktop E2E was intentionally not run locally; CI owns the virtual-display run and screenshot
- CI Web E2E passes, including the signed GitHub trigger flow and screenshot

The unrestricted all-repository unit command was also attempted locally, but unrelated suites that open local servers timed out under the restricted network/IPC environment; the affected and adjacent suites above pass.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub event triggers for routines, including creation, editing, validation, and status display.
  * Added signed GitHub webhook endpoints with sanitized event metadata and duplicate-delivery protection.
  * Routine trigger summaries now distinguish scheduled, webhook, and GitHub event triggers.
  * Added Chinese translations for GitHub event and webhook labels.

* **Bug Fixes**
  * Improved protection against untrusted webhook content and event-authored prompt instructions.
  * Webhook-triggered actions with side effects now require approval before execution.

* **Tests**
  * Added coverage for GitHub triggers, webhook security, validation, delivery handling, and approval safeguards.

### DIFF
--- a/apps/api/src/github-webhook.ts
+++ b/apps/api/src/github-webhook.ts
@@ -1,0 +1,157 @@
+import { createHmac, timingSafeEqual } from "node:crypto";
+
+export function hasValidGithubSignature(
+  signature: string | undefined,
+  secret: string,
+  raw: string,
+): boolean {
+  if (!signature || !/^sha256=[0-9a-f]{64}$/.test(signature)) return false;
+  const expected = `sha256=${createHmac("sha256", secret).update(raw).digest("hex")}`;
+  return timingSafeEqual(Buffer.from(signature), Buffer.from(expected));
+}
+
+const GITHUB_EVENT_NAMES = new Set([
+  "check_run",
+  "check_suite",
+  "create",
+  "delete",
+  "deployment",
+  "deployment_status",
+  "discussion",
+  "discussion_comment",
+  "issue_comment",
+  "issues",
+  "merge_group",
+  "ping",
+  "pull_request",
+  "pull_request_review",
+  "pull_request_review_comment",
+  "push",
+  "release",
+  "repository",
+  "repository_dispatch",
+  "status",
+  "workflow_dispatch",
+  "workflow_job",
+  "workflow_run",
+]);
+
+const GITHUB_ACTIONS = new Set([
+  "assigned",
+  "closed",
+  "completed",
+  "created",
+  "deleted",
+  "edited",
+  "in_progress",
+  "labeled",
+  "locked",
+  "opened",
+  "published",
+  "queued",
+  "ready_for_review",
+  "reopened",
+  "requested",
+  "review_requested",
+  "synchronize",
+  "unassigned",
+  "unlabeled",
+  "unlocked",
+  "unpublished",
+]);
+
+function record(value: unknown): Record<string, unknown> | undefined {
+  return value && typeof value === "object" && !Array.isArray(value)
+    ? (value as Record<string, unknown>)
+    : undefined;
+}
+
+export function githubEventName(header: string | undefined): string {
+  const value = header?.trim() ?? "";
+  return GITHUB_EVENT_NAMES.has(value) ? value : "event";
+}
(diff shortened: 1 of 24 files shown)


fetw882:
Addressed the prompt-injection finding in `5fec0a16`. GitHub deliveries no longer forward titles, bodies, comments, commit messages, logins, repository names, or other event-authored strings into the wake-up prompt. The prompt now contains only allowlisted event/action enums, numeric IDs, booleans, and validated commit hashes; regression coverage proves malicious free text is absent. The agent must fetch any details through its configured tools, where external content remains data rather than the initiating instruction.

fetw882:
The remaining review findings are addressed in `68c00a3f` and `9c0e1858`: oversized bodies are rejected before target lookup, per-bot secret provisioning is single-flight across trigger setup and Save, and plain-text or malformed-JSON generic webhook bodies are now fenced as untrusted delivery data instead of becoming direct agent instructions. Regression coverage passes in the API suite (253 tests) and web suite (204 tests); all 22 workspace typecheck tasks and the web production build also pass.

fetw882:
The advisory-only fence concern is now addressed structurally in `acefa2a1`. Runs with trigger `webhook` may execute read-only inspection unattended, but local side effects and mutating connector calls always pause for owner approval. This includes `shell` and `write_file`, and an existing `always_allow` rule cannot bypass the boundary. Integration regressions exercise both builtins and direct/catalog connector writes; 82 focused tests and all 22 workspace typecheck tasks pass.

PR #752, opened 2026-09-07
Fix desktop startup collisions and leaked computer networks

elie222:
## Why

Local desktop startup can fail when development servers occupy its ports or Docker runs out of network address pools. Failed computer provisioning can contribute to exhaustion by leaving a newly allocated network behind.

## Changes

- Allocate bot networks after home validation. Roll back newly created containers and networks on provisioning failure without forcing removal of running containers or disconnecting active endpoints. Preserve existing and shared networks, and keep the existing computer if replacement-network allocation fails.
- Give the managed desktop stack a separate default web port, allocate its API host port dynamically, and retry web-port conflicts twice. Persist the selected loopback origin and use it consistently for authentication, health checks, setup, and reopening.
- Keep standalone Compose port defaults, and remove the unnecessary default network used by offline initialization services.

Error copy appears only after the relevant failure:

> Docker has no free network address pools. Remove unused Docker networks or expand Docker’s address pools, then retry.

This replaces the generic startup error because retrying without freeing address space cannot recover.

> Could not bind a local port after retrying. Retry to choose another port.

This replaces fixed-port advice after automatic recovery is exhausted. Removing it would leave no explanation for the stopped startup. Both messages are already disclosed only when needed.

## Validation

Regression tests were observed failing before the fixes. All 441 targeted desktop, supervisor, and Compose tests pass; opt-in Docker integration tests remain skipped. Repository-wide typecheck and lint pass, and the new Electron test code typechecks. Docker Compose configuration validates both standalone defaults and managed ports/auth origins without starting services.

CI Electron coverage passes for replacement-origin opening/saving, bounded port retries, and address-pool errors. The [CI screenshot artifact](https://github.com/elie222/rakazo/actions/runs/34069487996/artifacts/10000077047) contains `09-setup-address-pool-exhausted.png` and `10-setup-ports-unavailable.png`; both were visually checked.

### DIFF
--- a/apps/desktop/e2e/local-stack.spec.ts
+++ b/apps/desktop/e2e/local-stack.spec.ts
@@ -119,7 +119,7 @@ async function writeFakeDocker(mode: FakeDockerMode) {
 }
 
 async function launch(mode: FakeDockerMode | "missing") {
-  const env = { ...process.env, RAKAZO_PERFORMANCE_USER_DATA: userData };
+  const env: NodeJS.ProcessEnv = { ...process.env, RAKAZO_PERFORMANCE_USER_DATA: userData };
   // A stale RAKAZO_WEB_URL from the developer's shell would bypass setup entirely.
   delete env.RAKAZO_WEB_URL;
   return electron.launch({
--- a/apps/desktop/e2e/local-stack.spec.ts
+++ b/apps/desktop/e2e/local-stack.spec.ts
@@ -1,5 +1,5 @@
 import { mkdir, mkdtemp, readFile, realpath, rm, stat, writeFile } from "node:fs/promises";
-import { createServer, type Server } from "node:http";
+import { createServer, type RequestListener, type Server } from "node:http";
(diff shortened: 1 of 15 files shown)


elie222:
@coderabbitai The non-forced rollback is intentional. A rejected start response can mean Docker started the managed computer but the response was lost. Forcing removal would destroy that running computer; its attached network is in use, not an orphaned empty network. A subsequent provision request finds the labeled container and reuses it.

The regression `preserves the provision error if Docker refuses cleanup of active resources` covers this case and asserts non-forced removal plus preservation of the original error. Separate tests cover removal of a stopped newly created container before its new network, repeated failed creation without accumulating networks, and preservation of an existing computer if replacement-network allocation fails. We should retain these safety guarantees rather than force removal. Please reassess the outside-diff finding on that basis.

The docstring percentage warning does not identify a missing behavioral contract. The changed cleanup and port-allocation code documents its non-obvious safety constraints; adding comments that repeat test helpers would not improve this fix.

Issue #816, opened 2026-09-08
[Bug]: No limit on how many computers a single user can create


ARMeeru:
### Confirmation

- [x] This is not a security vulnerability report (robustness gap)

### Description

Per-computer limits exist (`computerResourceLimits`, `infra/sandboxes/supervisor/src/computer-spec.ts:102-123`: memory, CPU, pids), but I can't find a cap on the count. The creation chain is authed oRPC → `DockerSandboxProvider.provision` (`packages/adapters/src/docker-sandbox.ts:136-162`) → supervisor `POST /computers`, and nothing in that chain checks how many computers a user or space already has. `SANDBOX_TEAM_SCREEN_LIMIT` caps screens only.

So one user with a loop, or a bot whose tool use loops, can create containers until the host runs out. Each container is individually bounded; the aggregate isn't. Low impact for a single-owner local deployment, more real for multi-user ones.

### Steps to reproduce

1. Start the stack and create a user
2. Create bots in a loop (API or UI); each provision provisions another computer container
3. Host resources exhaust while every individual container stays within its own limits

### Expected behavior

Some ceiling on computers per user, per space, or per deployment, checked at provision time.

### Actual behavior

No ceiling found. One user can create containers until the host runs out.

### Rakazo version or commit

main @ c288959

### Operating system

macOS

### SANDBOX_PROVIDER

docker

### AGENT_RUNTIME

pi

### Notes

One honesty note: this is an absence claim from searching `apps/api`, `packages/db`, `packages/adapters`, and the supervisor source. I traced the Docker path end to end; a limit inside the remote providers (E2B, Daytona, Box) I can't rule out.

Searched first; #770 is per-connection image limits, not computer counts. I'd suggest an env knob along the lines of `SANDBOX_MAX_COMPUTERS_PER_USER` checked at provision time, but the right scope (user, space, or deployment-wide default) is the maintainer's call. Happy to send a PR once the scope is settled.

PR #827, opened 2026-09-09
feat(computers): cap Computer records per user with SANDBOX_MAX_COMPUTERS_PER_USER

ARMeeru:
## Why

Nothing bounds how many Computer records one user can create. Every bot create, duplicate, and setComputer call funnels into `ensureComputerRecord`, and each row becomes a sandbox container once provisioned (default 2GB / 2 CPU per computer). The per-computer limits exist, but one user can keep creating records until the host runs out. Issue #816.

## What changed

- New env knob `SANDBOX_MAX_COMPUTERS_PER_USER`, resolved in `packages/db/src/computers.ts` following the `SANDBOX_TEAM_SCREEN_LIMIT` pattern. Unset, empty, or `0` means no configured cap (default, so existing deployments are unaffected).
- `ensureComputerRecord` counts the user's Computer rows before creating a new one and throws `ComputerLimitError` at the cap. The count only includes rows still referenced by a non-archived bot: archiving keeps the row, but the sandbox is released, so archived bots must not consume quota. A team computer is one row per space, count once, and reusing it never trips the cap.
- `mapSpaceLifecycleError` now surfaces `ComputerLimitError` as `BAD_REQUEST`, and the two `setBotComputer` call sites that create records are wrapped so the switch path returns the same error instead of a 500.
- `.env.example` documents the new knob next to `SANDBOX_TEAM_SCREEN_LIMIT`.

## How it was tested

- `packages/db/src/computers.test.ts` covers the resolver (unset/empty/0, positive integers, invalid values) and the three enforcement cases: new record at the cap throws, archived bots do not count, team computer reuse at the cap still succeeds.
- Verified red against today's code (3 failing tests for the resolver and enforcement), then green with the fix. `pnpm --filter @rakazo/db test`: 89 passed, 6 skipped (Postgres journeys); `@rakazo/db` and `@rakazo/api` typecheck clean; Biome lint and format clean on changed files.

## Out of scope

The supervisor and remote providers (E2B, Daytona, Box) are unchanged; this caps record creation at the DB chokepoint, which covers every provider. Runtime concurrency between two simultaneous creates is not fenced: two requests can each pass the count before either writes. A per-user lock at create time would close that gap, but it changes the create transaction and is worth its own discussion.

## Summary by CodeRabbit

- **New Features**
  - Added an optional per-user limit for live sandbox computers.
  - Existing computer records can be reused without consuming additional capacity.
  - A value of zero or an unset limit leaves computer creation unrestricted.
  - Restoring an archived bot now respects the configured computer limit.
  - Concurrent computer creation now enforces the limit reliably.

- **Bug Fixes**
  - Computer-capacity and sandbox lifecycle errors now return clear client errors consistently.

- **Tests**
  - Added coverage for configuration, enforcement, concurrency, reuse, restoration, and unrestricted operation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,6 +74,10 @@ SANDBOX_COMMAND_TIMEOUT_MS=300000
 # Optional limit on active bot desktops per Docker Team Computer.
 # 0 leaves capacity to the machine; desktops and Chrome start only when needed.
 SANDBOX_TEAM_SCREEN_LIMIT=0
+# Optional cap on Computer records still backing live bots, per user. Each record
+# becomes a sandbox container once provisioned, so the cap bounds the container
+# fleet one user can create. 0 means no configured cap (default).
+SANDBOX_MAX_COMPUTERS_PER_USER=0
 # Optional per-turn tool-call fuse for the Pi agent runtime. Unset, empty, or 0
 # means unlimited (default). Set a positive integer to soft-stop a turn that
 # exceeds the budget and still emit a final assistant message.
--- a/apps/api/src/router.test.ts
+++ b/apps/api/src/router.test.ts
@@ -4,7 +4,7 @@ import type { Actor } from "@rakazo/contracts";
 import { openScreenCapability } from "@rakazo/core/node/screen-capability";
 import type { PrismaClient } from "@rakazo/db";
 import { createLogger, createTestSink, installLogger } from "@rakazo/logging";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { createRouter, type RouterDeps } from "./router.js";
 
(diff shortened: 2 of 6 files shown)


ARMeeru:
Finding 1 is fixed in c391e4c: restore now enforces the quota the same way creation does, with regression coverage for the archive → create → restore sequence at both the db helper and router level.

Finding 2 is left for the maintainer by design. Serializing the check and create needs a per-user transaction lock, and the repo already has two candidate mechanisms: `lockSpaceForContentCreation` (space level) and `withTransactionRetry` (transaction level). Which one is right depends on how strict the cap should be, so that decision is better made here than inside this PR, in the same spirit as the Out of scope note above. Happy to either open a follow-up or land it in this PR if there's a preference.

PR #854, opened 2026-09-12
fix(adapters): preserve nullable connector tool parameters

fetw882:
## Why

Connector tools can describe nullable filters, cursors, and array elements with JSON Schema type arrays or explicit null branches. The runtime currently turns `type: ["object", "null"]` into a string schema, losing the object's fields, and turns `type: "null"` into a string as well. The model is then asked to send values that disagree with the connector's original argument validator.

This complements the top-level object envelope fix in #853: nested nullable fields still need to retain their actual types.

## What changed

- Translate JSON Schema type arrays into TypeBox unions using the existing field conversion.
- Preserve explicit null fields with `Type.Null()`.
- Cover nullable object fields, null union branches, nullable array elements, and the model-facing `parametersFor` boundary against connector argument validation.

No provider-specific switches, UI changes, or changes to connector execution policy.

## Testing

- Three wire-schema regressions fail on the original implementation and pass with the fix.
- Additional boundary regression verifies boolean/null values remain valid while strings and missing required values are rejected.
- Focused schema suites: 33 tests passed, including the additional boundary regression.
- Full adapters suite on the final commit: 1,667 passed, 25 skipped.
- Core suite: 458 passed.
- Adapters TypeScript check, targeted Biome check, and diff whitespace check passed.

No live model or external MCP server was used; the reproduction is deterministic and offline.

### DIFF
--- a/packages/adapters/src/pi-runtime-tool-schema.test.ts
+++ b/packages/adapters/src/pi-runtime-tool-schema.test.ts
@@ -1,7 +1,84 @@
 import { describe, expect, it } from "vitest";
-import { jsonSchemaParameters } from "./pi-runtime.js";
+import { parseConnectorToolArgs } from "./lazy-tool-catalog.js";
+import { jsonSchemaParameters, parametersFor } from "./pi-runtime.js";
 
 describe("jsonSchemaParameters", () => {
+  it("keeps model-facing nullable parameters compatible with connector validation", () => {
+    const tool = {
+      name: "catalog_lookup",
+      description: "Look up catalog entries",
+      inputSchema: {
+        type: "object",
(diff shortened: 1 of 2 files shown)


fetw882:
Re: https://github.com/elie222/rakazo/pull/854#issuecomment-5642962631 — Thanks for the review. Keeping this fix scoped to nullable schema fidelity and its regressions; the docstring-coverage warning concerns the existing jsonField function and does not identify missing behavior or validation. No documentation-only expansion is included. The final commit passes the complete adapters suite (1,667 tests; 25 skipped), core suite (458 tests), adapters typecheck, and targeted lint. Repository checks are terminal and passing.

PR #858, opened 2026-09-13
fix(adapters): respect selected subagent image limits

fetw882:
## Why

A subagent can select a different model connection with its own image limit, but its context transform still reads the parent model's limit. A stricter child can receive too many screenshots, while a more capable child can lose screenshots unnecessarily. A child with no configured limit also incorrectly inherits an explicit parent limit instead of the normal screenshot fallback.

## What changed

- Use the resolved subagent request model's `maxImagesPerPrompt` for the child context transform.
- Exercise the transforms installed by the actual runtime with parent/child limits of 2→1, 1→2, 2→0, and 0→unset. Assert that the parent retains its own budget in every case.

The screenshot pruning algorithm, model selection, and parent configuration are unchanged. This follows the per-connection image limits introduced in #761 and the existing per-call subagent model selection.

## Testing

- All four new cases fail on the original implementation and pass after the one-line fix.
- Focused runtime/computer tests: 22 passed.
- Full adapters suite: 1,671 passed, 25 skipped.
- Adapters TypeScript, targeted Biome and diff checks passed.
- No live provider or computer was provisioned; tests use the existing fake agent with the real context transform.

## Summary by CodeRabbit

- **Bug Fixes**
  - Screenshot context is now limited according to the model handling each subagent request, improving compatibility when parent and subagent models support different image limits.
  - Excess screenshot images are truncated consistently for both parent agents and subagents, helping prevent oversized prompts and related processing issues.

- **Tests**
  - Added coverage verifying image-limit behavior across parent and subagent execution scenarios.

### DIFF
--- a/packages/adapters/src/pi-runtime-thinking-level.test.ts
+++ b/packages/adapters/src/pi-runtime-thinking-level.test.ts
@@ -1,7 +1,9 @@
+import type { AgentMessage } from "@earendil-works/pi-agent-core";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const fakeAgentState = vi.hoisted(() => ({
   thinkingLevels: [] as string[],
+  transforms: [] as Array<(messages: AgentMessage[]) => Promise<AgentMessage[]>>,
   models: [] as Array<{
     id: string;
     provider: string;
@@ -28,6 +30,7 @@ vi.mock("@earendil-works/pi-agent-core", () => ({
 
     constructor(options: {
       sessionId?: string;
+      transformContext: (messages: AgentMessage[]) => Promise<AgentMessage[]>;
       initialState: {
         thinkingLevel: string;
         tools: FakeAgentTool[];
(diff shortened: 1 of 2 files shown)


elie222 (approved):
Clean one-line fix: subagent context pruning now uses the resolved child model's `maxImagesPerPrompt` instead of the parent's. Tests cover stricter/looser/zero/unset child budgets while keeping the parent budget intact. No UI change.

fetw882:
Thanks for the approval. The remaining Web E2E failure is in knowledge-panel.spec.ts:174: after saving and reopening a skill, the editor textarea was not found within 20 seconds. The run passed 112 other cases. The new subagent budget regressions and the complete adapters suite pass locally. I attempted to rerun the failed job, but GitHub requires repository admin rights. Could a maintainer rerun the failed Web E2E job? I have not established a base-branch reproduction, so I am not marking the failure unrelated or resolved. Run: https://github.com/elie222/rakazo/actions/runs/34729691227

PR #861, opened 2026-09-14
fix(web): disable skill rows while refreshing after save

fetw882:
## Why

After saving a skill, the editor closes before the catalog refresh finishes. Skill rows reappear while `busy` is still true, but look enabled even though `openSkill` ignores clicks during that interval. A quick click is silently lost; browser automation can also click the row and then time out waiting for an editor that never opened.

This matches the save/reopen failure observed in the knowledge-panel journey on #858. That run passed on retry, so a deterministic delayed-refresh regression is included rather than relying on the intermittent failure alone.

## What changed

- Disable skill rows while a mutation and its catalog refresh are pending, matching the existing New skill control and click guard.
- Add a component regression with a controlled refresh promise: rows are disabled during refresh and reopen normally afterward.
- Extend the existing knowledge-panel E2E save/reopen journey with a gated catalog refresh and a screenshot of the disabled row.

No new UI copy. Applies to web and the Electron-hosted web settings; native mobile is unchanged.

## Testing

- The new component regression fails on the original row and passes with the fix.
- Web unit suite: 212 tests passed.
- Web TypeScript, targeted Biome, and diff checks passed.
- Full browser journey runs in CI; not run locally because the integration database/container harness is unavailable.
- Web E2E passed in CI. [Pending-refresh screenshot](https://rakazogithubactions.fsn1.your-objectstorage.com/playwright/runs/34795159647-1/screenshots/images/100-85-skill-refresh-pending.png).
- Postgres journeys is not green: the group-chat headline-run assertion at `packages/testkit/src/journeys.test.ts:1892` failed. No base-branch reproduction has established whether this is independent of the PR.


elie222 (approved):
Clean one-line fix: skill rows now respect `busy` during the post-save catalog refresh, matching New skill and the existing `openSkill` guard. Unit test with a gated refresh promise and the knowledge-panel E2E screenshot of the disabled row look solid. No copy changes.

fetw882:
Web E2E passed, and the PR description now links the pending-refresh screenshot. The remaining failing check is [Postgres journeys](https://github.com/elie222/rakazo/actions/runs/34795159647/job/103826693969): 34 passed, one failed in the group-chat journey at packages/testkit/src/journeys.test.ts:1892, where the headline run differs from the waiting-input run. This was an actual assertion failure, not a setup failure; independence from this PR has not been established by a base reproduction. Could a maintainer rerun the failed job? The rerun request requires repository admin rights.

PR #885, opened 2026-09-16
fix(mcp): preserve discovery timeout with caller cancellation

fetw882:
## Why

MCP OAuth discovery can advertise an unreachable host while serving the same discovery document on the configured endpoint origin. The fallback helper intends to cap that first attempt at four seconds, but passing a caller AbortSignal currently disables the cap entirely. Discovery can consume the whole connection deadline before fallback gets a chance. Caller cancellation also enters the fallback path and attempts another fetch with an already-aborted signal.

## What changed

- Combine the existing four-second first-attempt deadline with the caller signal.
- Keep the original caller signal for fallback so the expired per-attempt deadline does not cancel the retry.
- Propagate caller cancellation without attempting fallback.
- Add deterministic regressions with injected fetches and a controlled deadline; no external requests or real timers are needed.

Same-origin, POST, Request-input, header sanitization and host policies are unchanged. This is separate from #881's private-endpoint policy: its edits in the same module do not touch the fallback helper. No UI changes.

## Validation

- Both regressions fail against the original code and pass fixed.
- Focused MCP transport/OAuth/fallback suites: 48 passed.
- Full adapters suite: 1,758 passed, 25 skipped.
- Adapters TypeScript, targeted Biome and diff whitespace checks passed.
- Frozen offline filtered dependency install succeeded on the current upstream baseline.
- No live provider or production database was used. Browser and database integration journeys are left to CI.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection setup reliability by limiting initial discovery attempts to four seconds.
  - Caller cancellation is now respected throughout connection setup, preventing unnecessary fallback attempts after an operation has been aborted.
  - Preserved the original cancellation behavior for subsequent connection requests.


fetw882:
Reviewed the [Greptile assessment](https://github.com/elie222/rakazo/pull/885#issuecomment-5692204087); no actionable findings remain. All code checks, including Postgres journeys and Web E2E, now pass on ea1f5cfb. The caller-cancellation and per-attempt timeout regressions pass without external requests.

PR #893, opened 2026-09-17
State the current date and time in every run's system instructions

muurlingjunior:
## Why

Bots currently receive no notion of the present moment. The system instructions built in `executor.ts` carry the bot's own instructions, workspace and tool guidance, but never the date or time, and neither does the Pi runtime underneath. The model therefore infers "now" from its training data or from whatever timestamps happen to appear in the conversation.

In practice this shows up as a bot that treats last week's items as today's, computes deadlines from a stale date, or writes "next Tuesday" against the wrong week. On a self-hosted deployment with routines and a persistent memory this is one of the first things a user notices.

## What

- New `formatCurrentTimeInstruction()` in `packages/adapters/src/current-time.ts`: one line stating the weekday and an ISO 8601 UTC timestamp (seconds precision), plus guidance to treat it as the present, convert to the user's time zone when known, write absolute dates, and never infer today's date from training data or quoted timestamps.
- `executor.ts` prepends that line to every run's system instructions, directly after the bot's own instructions.
- Unit tests for the formatter (fixed date, guidance text, default clock).
- CHANGELOG entry under Unreleased.

UTC is deliberate: Rakazo has no user or space time zone today (only routines carry one), and a wrong local time is worse than a correct UTC anchor. Layering a user time zone on top later is a one-line change in the formatter.

## Testing

- `pnpm --filter @rakazo/adapters test -- current-time` passes.
- `pnpm --filter @rakazo/adapters check` (tsc) passes.
- `pnpm lint` (Biome) passes on the touched files.
- Verified on a self-hosted instance that a bot asked "what day is it?" answers with the real date after this change.

## Summary by CodeRabbit

* **Changed**
  * Each run now includes the current date and time in UTC.
  * Responses can more accurately evaluate deadlines, recency, and scheduling.
  * Date-sensitive guidance now uses absolute dates, accounts for the user’s known time zone, and avoids relying on training data or quoted timestamps.


elie222:
Superseded by a maintainer branch that keeps this work and wires the current-time instruction into auto-review and history-compaction as well (fork was not pushable). See the new PR.

PR #897, opened 2026-09-17
Rename lazy MCP catalog wrappers: "mcp_" tool names are rejected by Claude Code OAuth

muurlingjunior:
## Symptom

A bot with more than 20 MCP tools fails on **every** Claude model behind a Claude Pro/Max/Team sign-in, immediately, with:

```
400 {"type":"invalid_request_error","message":"You're out of extra usage. Add more at claude.ai/admin-settings/usage and keep going."}
```

The account is nowhere near its limits, and a bot in the same space with 19 MCP tools runs fine on the same credential and model. The message points at billing, so it is easy to chase the wrong thing (we did, for an hour).

## Root cause

I captured the exact request Rakazo sends and replayed it against `api.anthropic.com` with single elements removed. Removing the tools made it pass; bisecting the 48 tools isolated three: `mcp_search_tools`, `mcp_load_tool`, `mcp_execute_tool`. A lone dummy tool named `mcp_x` reproduces the 400 on Sonnet 5, Opus 5 and Fable 5.1; `mcpx` and `mcp__server__tool` are accepted.

So Anthropic's Claude Code OAuth path rejects any tool whose name starts with `mcp_` (single underscore). The direct MCP tool names Rakazo uses (`mcp__server__tool`, double underscore) are fine; only the lazy catalog wrappers that kick in above `DIRECT_TOOL_LIMIT` hit it.

## Fix

- `McpConnector.discoverTools` now builds the wrappers with prefix `connectors`: `connectors_search_tools`, `connectors_load_tool`, `connectors_execute_tool`. The connector id stays `mcp`, and approval replay matches on the `_execute_tool` suffix, so routing and approvals are unchanged.
- Doc comment on `lazyCatalogTools` recording the constraint, plus a regression assertion that no wrapper name starts with `mcp_`.
- Tests updated; CHANGELOG entry under Fixed.

## Testing

- `pnpm exec vitest run` on `mcp-connector`, `lazy-tool-catalog`, `executor-approval-replay`, `approval-effect`: 70 tests pass.
- `pnpm --filter @rakazo/adapters check` and Biome pass.
- On our self-hosted instance (Team org, overage disabled) the failing bot completes its runs on Opus 5 once the wrapper names avoid the `mcp_` prefix, verified by replaying the captured request with the renamed tools.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where bots with more than 20 MCP tools could fail on Claude models using Claude Pro, Max, or Team sign-in with an “out of extra usage” error.
  - Connector catalog tools now use `connectors_*` names.
  - Directly exposed MCP tools remain available with their existing names.
  - Approval flows continue to work with the updated connector catalog tool names.

- **Documentation**
  - Added release notes describing the compatibility fix and tool naming requirements.


elie222 (changes requested):
Thanks for the clear root-cause writeup and the wrapper rename — the Anthropic `mcp_` rejection is real and the `connectors_*` names look right for discovery.

One gap before this can land: `catalogExecuteToolName(connectorId)` still builds `` `${connectorId}_execute_tool` ``, so for MCP it returns `mcp_execute_tool` while `discoverTools` now exposes `connectors_execute_tool`. That breaks `buildApprovalContinuation` when a bound direct MCP approval must resume after the catalog grows past the direct limit (`exposed.has(wrapper)` misses). Same mismatch: `catalogApprovalConnectorId("connectors_execute_tool")` yields `connectors` instead of `mcp`.

Please add a small prefix helper (`mcp` → `connectors`, otherwise identity), use it from `catalogExecuteToolName` and from the MCP `lazyCatalogTools` call, map `connectors` back to `mcp` in `catalogApprovalConnectorId`, and add a regression test like the installed `afterGrowth` case for MCP. Optional but useful: treat a legacy `mcp_execute_tool` envelope as `connectors_execute_tool` when the new wrapper is exposed so in-flight approvals still resume.

No UI here, so no E2E screenshot needed.

elie222 (approved):
Follow-up `b0644f01` covers the gap from the earlier review: shared `catalogToolPrefix`, bound-MCP afterGrowth resume, and legacy `mcp_execute_tool` alias. Greptile P1 is resolved. Required CI is green. No UI, so no E2E frame needed.

PR #901, opened 2026-09-18
fix(mcp): replay legacy approvals through renamed catalog wrapper

fetw882:
## Why

After #897, an approval saved with `mcp_execute_tool` is resumed with instructions to call `connectors_execute_tool`. The replay gate still compares the wrapper names literally, rejecting the instructed call before catalog resolution. A previously approved action can therefore remain stuck after upgrading.

## What changed

- Accept the specific legacy MCP wrapper-to-current-wrapper transition at the existing catalog replay gate.
- Keep exact matching for all other wrappers and retain the non-catalog guard. Reuse the persisted approved arguments; resource and revision validation are unchanged.
- Extend the existing continuation test through replay selection, catalog resolution, resource checks, argument restoration and queue draining. Cover mismatched wrappers, reverse aliasing, changed resources/revisions and non-catalog calls.

No schema migration, frontend, copy, permission-policy or provider changes.

## Validation

- The expanded legacy replay regression fails on the base with the old-wrapper-before-new-wrapper error, then passes fixed.
- Full adapters suite: 1,794 passed, 25 skipped. Tests use scripted/fake fixtures without live provider credentials or production database access.
- After correcting a test-only type narrowing, the focused replay suite passes all 18 tests and adapters TypeScript passes.
- Targeted Biome and diff whitespace checks pass.
- This validates the existing replay-helper chain, not a live external MCP side effect. Full repository CI follows on the PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Approval replays now continue to work when an older MCP tool wrapper has been replaced by its renamed catalog execution tool.
  * Unrelated tool wrappers remain protected from incorrect approval reuse.
  * Improved handling of resource revisions, arguments, successful resolutions, and replay errors.


fetw882:
Follow-up to https://github.com/elie222/rakazo/pull/901#issuecomment-5723467333 : confirmed on head 3d0f7eef that all repository checks are terminal (12 successful, 4 skipped), with both review checks complete. The regression fails before the fix and passes afterward; local adapters tests report 1,794 passed and 25 skipped, with TypeScript and Biome clean. The alias remains one-way and does not change resource/revision validation or the approved arguments.

PR #926, opened 2026-09-18
fix(adapters): keep MCP catalog wrappers off the mcp_ tool-name prefix

shivendrasoni:
## Why

A bot with more than twenty MCP tools could not run at all. Even a plain
"hi" came back as:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"You're out of extra usage. Ask your workspace admin to add more so you can keep going."}}
```

The message points at the workspace's usage allowance, so it sends you to
your billing admin. That is not the cause. Anthropic rejects any request
whose tool list contains `mcp_`-prefixed tool names, and reports it with
that billing text instead of a naming error.

Above `DIRECT_TOOL_LIMIT` the MCP connector replaces the server's tools
with three catalog wrappers, which were named `mcp_search_tools`,
`mcp_load_tool` and `mcp_execute_tool` — so crossing the limit broke the
bot entirely, on every message, with an error that named the wrong thing.

## What changed

The wrappers use the `toolkit_` prefix. Nothing else about the catalog
changes: same schemas, same routes, same search/load/execute flow.

The reason is recorded next to the call and asserted in a test, because
the obvious cleanup is to rename them back to match the connector id.

## How it was tested

Reproduced against a live MCP server with 21 tools. The same bot, in the
same thread, on the same credential and model, seconds apart: the request
carrying the `mcp_*` wrappers was rejected, and the one carrying
`toolkit_*` succeeded. The failing request was the smaller of the two
(29,032 bytes of tool definitions against 27,833 for a working request
with more history), so payload size was not the cause. Search, load and
execute were then exercised end to end against that server.

Unit tests updated across the three affected files; all pass.

## Not covered

Tools on the direct path below the limit are still named
`mcp__<slug>__<tool>`, which also begins with `mcp_`. Whether the double
underscore form trips the same rejection is untested, so bots under the
limit may hit this too. That path is unchanged here.

## Summary by CodeRabbit

* **Bug Fixes**
  * MCP-discovered tools now use `toolkit_`-prefixed wrapper names instead of `mcp_`-prefixed names.
  * Prevents compatibility errors with Anthropic requests involving MCP tools.
  * Approval and continuation flows now consistently reference the updated wrapper names.


elie222:
Thanks for the report and the fix — same root cause already landed on main in #897 (catalog wrappers use `catalogToolPrefix(\"mcp\")` → `connectors_*` so they stay off the `mcp_` prefix Anthropic rejects). Closing this as superseded by that merge.

PR #942, opened 2026-09-20
Add an Artifacts tab: browse, preview, version, and manage what bots make

paulpreibisch:
## Why

I wanted our bots to have the ability to publish artifacts like Claude Code does, so we can see the output our bots create easily. Right now, anything a bot attaches is just a generic file chip buried in a chat thread — markdown gets a preview, everything else is a plain download, there's no way to browse everything a bot has made in one place, and re-attaching something with the same name just creates an unrelated duplicate instead of a new version.

## What this adds

- **A space-wide Artifacts tab** (`/app/artifacts`, `/app/artifacts/:id`), reached via a persistent left-edge rail (Bots / Artifacts) that's always visible, so switching between them never needs a back button. Card and list views, search, a date filter, and per-bot filter chips (colored/avatar'd to match each bot) live under a collapsible Filters toggle.
- **Split view**: selecting an artifact shows an index of everything on the left and the open artifact's preview on the right, with a maximize toggle to go full-width and back.
- **Inline preview** for HTML (sandboxed iframe — `sandbox="allow-scripts"`, no `allow-same-origin`, `srcdoc`, a locked-down CSP, so bot-authored HTML can run but can't touch the viewer's session or call back into the API), markdown, PDF (native browser viewer, no new dependency), and images. The existing inline chat attachment card gets the same treatment instead of only offering markdown preview or a download.
- **Versioning**: attaching something with the same name (same bot, same scope) is now a new version of the existing artifact instead of a duplicate. `Artifact` gains `rootArtifactId` (self-relation, null on a family's first version) + `version`; a family's stable identity is always its root's id. One shared resolver is used by both authoring paths (the `create` RPC and the `attach_file` tool) so the matching logic isn't duplicated. The detail view gets a version switcher (defaults to latest) and cards/rows show a version badge when a family has more than one.
- **Delete**: a hover-revealed control on every card/row opens a confirmation dialog and removes a whole family — every version's storage blob plus its DB rows (deleting the root cascades to its versions).
- Artifacts also gain an optional **description**, settable via `name`/`description` args on `attach_file` (nudged in its own tool description) or the `create` RPC, shown on cards so the list is scannable without opening anything.

## Test plan

- [x] `pnpm --filter <affected-packages> run check` (contracts, core, db, adapters, api, web) — clean
- [x] `pnpm exec biome check` on all touched files — clean
- [x] `pnpm --filter @rakazo/web run build` — succeeds
- [x] Full vitest suite (api, web, db, adapters, contracts, core) — passes; only pre-existing Windows-environment-specific failures (POSIX symlink/permission semantics, Linux shell script tests) unrelated to this change
- [x] Verified live against a real self-hosted instance (bot creating HTML/PDF/image artifacts, versioning via re-attach, delete, search/filter)

## Summary by CodeRabbit

* **New Features**
  * Added an Artifacts workspace with search, date and bot filters, grid/list views, pagination, version history, downloads, and deletion.
  * Added artifact descriptions, version information, and navigation between Bots and Artifacts.
  * Added HTML, PDF, and image previews, plus isolated HTML rendering.
  * File attachments now support HTML files, custom names, and descriptions.
* **Bug Fixes**
  * Improved artifact loading, version handling, cleanup, concurrent updates, and unsupported-file messaging.
  * Updated translations and user-facing messages.

## Screenshots

Illustrative — fabricated data throughout, no real bot names/content from any live instance.

**Browsing (card/list toggle, split view once something's open):**
Artifacts list view

**Filters open (search, date filter, per-bot chips):**
Filters open

**Bots view — the persistent rail (Bots / Artifacts) is what makes switching need no back button:**
Bots view


paulpreibisch:
Hey @elie222 — the Vercel check on this PR is showing a failure, but it looks like it's just the fork-authorization gate (Vercel needs someone on the Inbox Zero team to authorize preview deployments for PRs from a fork). Could you authorize it via the link on that check, or from the Vercel dashboard's pending deployments? All the actual CI checks (lint, typecheck, unit tests, Postgres journeys, Web E2E, production builds) are green. Also just pushed a merge with `main` to resolve the merge conflicts that had built up.

PR #947, opened 2026-09-20
fix(adapters): serve Pipedream tools through the lazy catalog

RoadJDK:
## Why

The Pipedream connector lists every tool of every connected app in one flat list, and all of them are offered on every run. `DIRECT_TOOL_LIMIT` is 20, so a single connected app can already pass it and a handful leave little room for anything else. That is the exact problem `lazy-tool-catalog.ts` already solves: above `DIRECT_TOOL_LIMIT` the MCP connector and the installed-connector source replace their tools with three wrappers the model searches, loads and executes through. Pipedream never opted in, so a handful of connected apps can fill a run's tool list on their own.

This closes one of two gaps. Composio still exposes its tools flat and is untouched here.

## What changed

`pipedream-connector.ts` only, following `McpConnector` line for line:

- The existing listing becomes a private `allTools`. `discoverTools` returns it unchanged at or below `DIRECT_TOOL_LIMIT`, and above it returns `lazyCatalogTools(...)`: `pipedream_search_tools`, `pipedream_load_tool`, `pipedream_execute_tool`.
- Each tool route carries `catalogGroup: app`, so catalog names group by app slug instead of collapsing into `other`. This is the one Pipedream-specific part: MCP groups by server slug, and the app slug is the equivalent identity here.
- `resolveCall` resolves a catalog execute id back to the owning tool, so the executor keeps parsed arguments and binds approvals to the real route.
- `execute` handles the three control routes before the app-route check, which would otherwise reject them with "Pipedream app route is missing".

The prefix comes from `catalogToolPrefix("pipedream")` rather than a literal. It is the identity today, but `catalogExecuteToolName` derives the approval-side name from the same helper, so the two stay one source of truth if that mapping ever grows. The catalog label is `app`, which is what the shared wrapper descriptions read correctly with.

Routes, execution path, redaction and error sanitizing are untouched. Nothing is added to the database, to the dependencies or to the interface. The CHANGELOG entry under Fixed names the three wrapper tools.

## How it was tested

New tests in the Pipedream connector suite, in the style of the MCP connector's catalog tests, against a fake fetch that serves the token endpoint plus one MCP server per app slug:

- Twenty tools stay direct; twenty-one switch to the three wrappers, and no tool schema reaches the wrapper definitions.
- An empty catalog search returns names grouped under each connected app slug.
- A catalog execute call for one app's tool resolves to that tool and reaches the MCP endpoint with that app's slug header and the parsed arguments — a generic catalog assertion would not prove the call landed on the right app.
- A direct app route is left to the normal execute path.

Each of the four fails against the unchanged connector and passes with the change.

Full Pipedream connector file: 17 tests pass. Full adapters suite: 1,853 passed, 25 skipped. Adapters TypeScript and Biome on the touched files are clean.

The same behaviour runs on a self-hosted instance with ten connected apps.

## Not covered

- Composio has the same flat listing and is not changed here.
- Every search, load or execute through a wrapper re-lists the tools of every connected app: one MCP initialize and `tools/list` per app, the same round trips `discoverTools` already makes at the start of a run. Ten connected apps means ten. No caching is added here.
- Tests use a fake fetch and fake credentials. No live Pipedream account is used in CI, so the catalog is not exercised against the real provider by these tests.

## Summary by CodeRabbit

* **New Features**
  * Added a lazy tool catalog when connected apps provide more than 20 tools.
  * Tools can be searched by app, loaded as needed, and executed through the catalog.
  * The catalog includes grouped app names and supports all connected apps.

* **Bug Fixes**
  * Large tool collections no longer expose every tool at once.
  * Failures discovering one app’s tools no longer prevent other apps from appearing.
  * Catalog executions are routed to the correct connected app and report failures safely.


elie222:
Addressed the CodeRabbit note on the 20-app discovery slice in d161651: every connected app is enumerated now, and a regression covers 21 one-tool apps switching to the lazy catalog. Docstring coverage nit skipped.

elie222:
Follow-up for the Greptile P1: discovery now catches per-app failures so one bad connection cannot hide the rest (same pattern as MCP).

